PR #12, opened 2022-07-26
update to dev api doc

thisab:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

**Description of change**
Added podman option in Makefile for api-doc

**Which issue this PR fixes**


pkg/server/server.go
```diff
@@ -5,12 +5,137 @@
...
 
 type Server struct {
-	// config Config
+	//config Config
 }
 
 func (s Server) Run() {
 	fmt.Println("Hello, world!")
 }
+
```
n-th, line 22: The content below the function Run should be separated from this file, as they are only handlers for the API server. You can remove them from here and add them to the handler.go file inside the proper API.

dev/api/harvester.html
```diff
@@ -0,0 +1,37 @@
...
+        main {
+            flex: 1 0 0;
+            overflow: hidden;
+        }
+    </style>
+</head>
+
+<body>
+    <main role="main">
+        <elements-api apiDescriptionUrl="oapi-codegen/harvester.yaml" router="hash" />
```
n-th, line 33: I would change this path to allow devs to run the server for the documentation locally instead of only relying on running this on Docker/Podman. For that, the URLs and file paths would need to be the same inside the container and out of it.

n-th (changes requested):
I think it is better to not push testing code to the master branch. I would remove that and add just the updates you have done to the dev API doc.

mchurichi:
hey @thisab, glad you got the gorm stuff working!, but I agree with Nath that we shouldn't be pushing testing stuff to the master branch, better to have this in a personal branch or a gist if we want this to be shared with the team, and at the same time, we'd rather have atomic PRs addressing a single thing at the time to facilitate reviews.

jufantozzi:
Just to line out something here: before this PR was opened, I asked @thisab offline to keep the code he used for testing temporarily in the PR so it would facilitate testing for us, as reviewers 😃.  So the intent of the testing code is in fact to be removed (as stated in coments in the code) once we were done testing.

With that said, i'll be reviewing this and running my tests ASAP👍

mgbcaio:
Good work, @thisab. I agree with Maxi and Nathalia regarding the atomic PRs and the test chunk of code. I think if we really need to share this code we could have it in a [gist](https://gist.github.com/). We would only add in the main branch to remove it later anyways. 👍

pkg/server/server.go
thisab: Ok, moved this code to handle.go

dev/api/harvester.html
thisab: Sure, but think doing "python3 -m http.server " on the same directory does not allow for accessing files on upper directories. I suggest looking into it later on.

cmd/server/main.go
```diff
@@ -21,8 +21,6 @@ import (
 )
 
 func main() {
-	var port = flag.Int("port", 8080, "Port for HTTP Galadriel server")
```
jufantozzi, line 24: Accidentally removed?

pkg/server/datastore/models.go
```diff
@@ -0,0 +1,27 @@
+package datastore
+
+import (
+	"gorm.io/gorm"
+)
+
+//TODO - Remember that the DeletedAt is removed in Spire
```
jufantozzi, line 7: Not that we should be worrying about this at this moment, but SPIRE doesn't seem to use `gorm.Model` for models that it doesn't want to be capable of doing deletions.

pkg/server/server.go
```diff
@@ -5,12 +5,6 @@
 // The code under api/petstore/ has been generated from that specification.
 package server
 
-import "fmt"
-
 type Server struct {
-	// config Config
-}
-
-func (s Server) Run() {
```
jufantozzi, line 14: accidental deletion?

pkg/server/server.go
```diff
@@ -5,12 +5,6 @@
 // The code under api/petstore/ has been generated from that specification.
```
jufantozzi, line 5: Could you please remove all these initial comments?

jufantozzi (commented):
Amazing work @thisab! 🏅 

Just left some minor comments and we should be good to go.

pkg/server/server.go
thisab: sure, removed on the last push

pkg/server/server.go
thisab: good call, readded the Run()

cmd/server/main.go
thisab: Sure, readded

jufantozzi (approved):
💯

pkg/server/datastore/models.go
thisab: Sure I agree.
Also, Ive spoken to Nathalia and we agreed on separating the DB code on another branch

n-th (approved):
🥓

n-th (dismissed):

mchurichi (dismissed):
🙌

n-th (approved):
👍

mchurichi (approved):

PR #16, opened 2022-07-29
Harvester boilerplate

mchurichi:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
N/A

**Description of change**
The idea of this PR is to start outlining the general modular architecture of the Harvester (the Server should be in the same line). All components have a very minimal interface and a weak implementation that will change as soon as we start tackling them one by one.

The components have been shaped into different packages with well-defined interfaces. A central component called Manager takes care of the initialization and hydrates the components catalog. Some of the components must run in the background, those implement the Runnable interface. The Controller (which is a Runnable) will query from the other components (brokers), pulling and pushing stuff as needed.

**Which issue this PR fixes**
N/A


n-th (approved):
🔥

n-th (dismissed):

pkg/common/log.go
```diff
@@ -0,0 +1,51 @@
+package common
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strings"
+)
+
+var (
```
mgbcaio, line 10: Wouldn't this be better to be a `const` ?

cmd/harvester/cli/cli.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+func (c *HarvesterCLI) Run(args []string) int {
+	if len(args) != 1 {
+		c.logger.Error("Unknown arguments", args)
+		return 1
+	}
+
+	cfg, err := config.LoadFromDisk("conf/harvester/harvester.conf")
```
mgbcaio, line 27: Maybe move this path inside a `const` ? We can also do it later :)

mgbcaio (approved):
Just left some comments. Good work, Maxi! 💯

pkg/common/log.go
mchurichi: good catch

mgbcaio (approved):
🦾

PR #20, opened 2022-08-01
DB update

thisab:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Added support for MySQL and PostgreSQL

**Description of change**

**Which issue this PR fixes**


mchurichi:
hey Thiago, any reason why you want to merge this into the `db` branch and not `master`? will you be using it as a feature branch?

thisab:
Hello Maximiliano, wont be a feature branch, was just a convention. What do you think @n-th ? Merge this into master?

n-th:
>  Hello Maximiliano, wont be a feature branch, was just a convention. What do you think @n-th ? Merge this into master? 
> > hey Thiago, any reason why you want to merge this into the db branch and not master? will you be using it as a feature branch?

The last PR related to db wasn't completed enough to be merged to the master branch, so we created this one to be an intermediate branch for the feature related to db in order to not pollute the main branch.

mchurichi:
> 

ok it makes sense, thanks to both

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,136 @@
...
+	return nil
+}
+
+func InsertOrg(db *gorm.DB, org Organization) error {
+	// Insert new ORG into the DB
+	err := db.Where(&org).FirstOrCreate(&org).Error
+	if err != nil {
+		fmt.Println("storage err: ", err)
+		return err
+	}
```
mchurichi, line 55: super nit: this can be simplified like this
```suggestion
	if err := db.Where(&org).FirstOrCreate(&org).Error; err != nil {
		return fmt.Errorf("sqlstorage error: %v", err)
	}
	return nil
```

mchurichi (commented):
Overall looks good to me 👌 
I know this is WIP and won't be merged into master now, but I left some minor comments to help start defining style and conventions over code in general.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,136 @@
...
+
+const (
+	// MySQL database type
+	MySQL = "mysql"
+	// PostgreSQL database type
+	PostgreSQL = "postgres"
+	// SQLite database type
+	SQLite = "sqlite3"
+)
+
```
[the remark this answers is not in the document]
thisab: Yes, I think the intent is really start outlining the datastore interface. Any more suggestions on that interface?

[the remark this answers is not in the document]
mchurichi: Cool. So, I think it shouldn't be tied to a specific implementation (this gorm-based plugin, for instance). Maybe a DataStore interface that defines the contract for data store plugins, with the basic CRUD operations using agnostic data models, and your sqlstore plugin should implement this interface.
Something similar to what we have already started defining on the harvester side: https://github.com/HewlettPackard/galadriel/blob/c23c4a6317ebf8ed0672ea23aadd188f3f1b0d4e/pkg/harvester/spire/spire.go

pkg/server/datastore/sqlstore/postgres.go
```diff
@@ -0,0 +1,17 @@
+package sqlstore
+
+import (
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+)
+
+type postgresDB struct{}
```
mgbcaio, line 8: nit: 
```suggestion
type PostgresDB struct{}
```

pkg/server/datastore/sqlstore/sqlite.go
```diff
@@ -0,0 +1,17 @@
+package sqlstore
+
+import (
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+type sqliteDB struct{}
```
mgbcaio, line 8: nit: 
```suggestion
type SqliteDB struct{}
```

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return nil
+}
+
+func CreateMemberTableInDB(db *gorm.DB) error {
+	// Creates the Model for the Bridge
+	result := db.AutoMigrate(&Member{})
+	if result != nil {
+		fmt.Println("storage err: ", result)
+		return result
```
mgbcaio, line 130:
```suggestion
	err := db.AutoMigrate(&Member{})
	if err != nil {
		fmt.Println("storage err: ", err)
		return err
```

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,133 @@
...
+func InsertOrg(db *gorm.DB, org Organization) error {
+	// Insert new ORG into the DB
+	if err := db.Where(&org).FirstOrCreate(&org).Error; err != nil {
+		return fmt.Errorf("sqlstorage error: %v", err)
+	}
+	return nil
+}
+
+func RetrieveOrg(db *gorm.DB, org *Organization) error {
+	// Insert new ORG into the DB
```
mgbcaio, line 57:
```suggestion
// RetrieveOrg retrieves an Organization from the Database. returns an error if something goes wrong.
func RetrieveOrg(db *gorm.DB, org *Organization) error {
```
Having this comment for the other functions will help as well. I advise you to put comments in the other functions. It provides a better description of the function even if it is called outside the file.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,133 @@
...
+	result := db.Where("Name = ?", (*org).Name).First(org)
+	if errors.Is(result.Error, gorm.ErrRecordNotFound) { // If does not, throw an error
+		fmt.Println("storage err: ", gorm.ErrRecordNotFound)
+		return errors.New("Organization " + (*org).Name + " does not exist in DB")
+	}
+	return nil
```
mgbcaio, line 63: shouldn't we be returning the result, aka Organization?

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		fmt.Println("storage err: ", err)
+		return err
+	}
+	return nil
+}
+
+func RetrieveBridge(db *gorm.DB, br *Bridge) error {
+
+	err := db.Where(br).First(br).Error
```
mgbcaio, line 91:
```suggestion
	result := db.Where(br).First(br)
```
and then use the `result.Error` for the error handling parts of the code.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		fmt.Println("storage err: ", err)
+		return err
+	}
+	return nil
```
mgbcaio, line 96: shouldn't we be returning the result from the DB transaction, aka `Bridge`?

pkg/server/datastore/sqlstore/models.go
```diff
@@ -29,7 +35,7 @@ type Bridge struct {
 }
 
 type Organization struct {
-	gorm.Model
+	Model
```
mgbcaio, line 38: Why use only `Model` instead of `gorm.Model` like the `Member`struct?

mgbcaio (changes requested):
Hey @thisab, left some comments! Let me know what you think. :)

pkg/server/datastore/sqlstore/sqlstore.go
thisab: It does, but it searches for the parameter inside the Organization passed as argument. This function was changed in a new code that is local in my machine to search for a string and return the Organization.

pkg/server/datastore/sqlstore/models.go
thisab: This is similar to the code in Spire. It removes the possibility of soft-deletion.

pkg/server/datastore/sqlstore/sqlstore.go
thisab: ok

pkg/server/datastore/sqlstore/mysql.go
```diff
@@ -0,0 +1,17 @@
+package sqlstore
+
+import (
+	"gorm.io/driver/mysql"
+	"gorm.io/gorm"
+)
+
+type mysqlDB struct{}
```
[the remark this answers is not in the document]
thisab: Why would this struct be public?
the code before is similar in structure to Spire: https://github.com/spiffe/spire/blob/main/pkg/server/datastore/sqlstore/mysql.go

pkg/server/datastore/sqlstore/dialect.go
```diff
@@ -0,0 +1,7 @@
+package sqlstore
+
+import "gorm.io/gorm"
+
+type dialect interface {
```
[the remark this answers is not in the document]
thisab: Why would it be public?
The code was also similar to Spire code: https://github.com/spiffe/spire/blob/main/pkg/server/datastore/sqlstore/dialect.go

pkg/server/datastore/sqlstore/sqlite.go
```diff
@@ -0,0 +1,17 @@
...
+
+func (sqliteDB) connect(connectionString string) (db *gorm.DB, err error) {
+	db, err = gorm.Open(sqlite.Open(connectionString), &gorm.Config{})
+	if err != nil {
+
+		return nil, err
+	}
```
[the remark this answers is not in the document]
thisab: ok

pkg/server/datastore/sqlstore/postgres.go
```diff
@@ -0,0 +1,17 @@
...
+
+func (postgresDB) connect(connectionString string) (db *gorm.DB, err error) {
+	db, err = gorm.Open(postgres.Open(connectionString), &gorm.Config{})
+	if err != nil {
+
+		return nil, err
+	}
```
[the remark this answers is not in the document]
thisab: ok

pkg/server/datastore/sqlstore/mysql.go
```diff
@@ -0,0 +1,17 @@
...
+
+func (mysqlDB) connect(connectionString string) (db *gorm.DB, err error) {
+	db, err = gorm.Open(mysql.Open(connectionString), &gorm.Config{})
+	if err != nil {
+
+		return nil, err
+	}
```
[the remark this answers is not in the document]
thisab: ok

pkg/server/datastore/sqlstore/postgres.go
thisab: I think this should be private as per other suggestions

pkg/server/datastore/sqlstore/dialect.go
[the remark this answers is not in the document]
mgbcaio: my bad, I thought you were using different packages. This leaves me wondering, will we be using this in another package?

pkg/server/datastore/sqlstore/sqlstore.go
mgbcaio: Oh, would you care to commit that to this PR so that it would be more consistent? thanks! 😊

pkg/server/datastore/sqlstore/models.go
mgbcaio: Got it, so the `Member` struct would also need to be `Model` instead of `gorm.Model` ?

pkg/server/datastore/sqlstore/dialect.go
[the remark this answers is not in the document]
thisab: Not outside the package, but by the function that opens DB using Gorm in a generic way (multiplexes into MySQL, Postgres or sqlite).

pkg/server/datastore/sqlstore/models.go
thisab: sure would, changed it

pkg/server/datastore/sqlstore/sqlite.go
thisab: I think this should be private as per other suggestions.

pkg/server/datastore/sqlstore/sqlstore.go
thisab: hum.. thinking about the creation workflow: the code that asks for the creation of, say Bridge, would have already created a Bridge "object" and passed it as an argument to this function. So the code that called the function would already have the it and might only be interested in knowing whether the operation succeeded.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,136 @@
...
+
+	switch dbtype {
+	case SQLite:
+		dialectvar = sqliteDB{}
+	case PostgreSQL:
+		dialectvar = postgresDB{}
+	case MySQL:
+		dialectvar = mysqlDB{}
+	default:
+		return nil, errors.New("unsupported database_type: %s" + dbtype)
```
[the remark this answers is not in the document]
thisab: sure, ok

thisab (commented):
Reviewed the suggestions and compiled them in the most recent commit

thisab (commented):
For this code I intend to add the implementation of the abstration layer (interface) later on. That way seems to make more sense as we define the API first. What do you think?

pkg/server/datastore/sqlstore/models.go
```diff
@@ -0,0 +1,69 @@
...
+	bridge        Bridge `gorm:"foreignKey:BridgeID;references:ID"`
+	MemberConsent bool
+	TTL           uint
+}
+
+type Relationship struct {
+	Model
+	MemberID            uint // Implicit Foreign Key and also the SourceID for the relationship
+	TargetMemberID      uint
+	SourceMemberConsent bool
```
jufantozzi, line 58: One thing that I was thinking about this is that "Source/Target" concepts makes sense if relationships are member initiates, but what if they were server initiated? Would Source/Target still make sense?

jufantozzi (commented):
Hey Thiago! thanks for this! Overall looking really good!

One thing that cought my attention is that we're starting off supporting muiltiple databases, and I could be wrong here, but my concern is that this could make it difficult to maintain further on, just adding extra overhead. 

Again, my lack of knowledge on gorm could mislead me here but IIRC we agreed on sqlite for the POC, and I'd probably vote for us to keep on that track, but would love you hear you back on this.

thisab:
> Hey Thiago! thanks for this! Overall looking really good!
> 
> One thing that cought my attention is that we're starting off supporting muiltiple databases, and I could be wrong here, but my concern is that this could make it difficult to maintain further on, just adding extra overhead.
> 
> Again, my lack of knowledge on gorm could mislead me here but IIRC we agreed on sqlite for the POC, and I'd probably vote for us to keep on that track, but would love you hear you back on this.

Regarding multiple DBs, there were no added complexity by using MySQL and Postgre. Just two files were needed and thats it. Basic functions are the same.

jufantozzi:
> Regarding multiple DBs, there were no added complexity by using MySQL and Postgre. Just two files were needed and thats it. Basic functions are the same.

Right, true for basic functions, but will this remain true when we start implementing the wanted behavior? My concern is that futurely we'd find out that there are many differences and this would pose unwanted difficulties...

pkg/server/datastore/sqlstore/models.go
thisab: If its server initiated, the status could change to reflect the server initiated approval. Source/Target does not have a meaning in that context, but its already there for that feature to be implemented later on.

thisab:
> > Regarding multiple DBs, there were no added complexity by using MySQL and Postgre. Just two files were needed and thats it. Basic functions are the same.
> 
> Right, true for basic functions, but will this remain true when we start implementing the wanted behavior? My concern is that futurely we'd find out that there are many differences and this would pose unwanted difficulties...

For the present, actually SQLite was a little harder because of manual cascading code. 
Cant say for the future. Gorm can even handle advanced queries well. Unless we start seeing bugs with gorm or start needing to handle raw queries.

For us to move on, then we can just delete the files mysql.go and postgres.go and thats it. No support for those two DBs. Agreed?

jufantozzi:
> For us to move on, then we can just delete the files mysql.go and postgres.go and thats it. No support for those two DBs. Agreed?

Since what he have here works, my suggestion is that we keep it, and see if in the future is still easy to support. If we find otherwise, then we can drop it. WDYT?

thisab:
> > For us to move on, then we can just delete the files mysql.go and postgres.go and thats it. No support for those two DBs. Agreed?
> 
> Since what he have here works, my suggestion is that we keep it, and see if in the future is still easy to support. If we find otherwise, then we can drop it. WDYT?

Agreed, as it is easy to remove support later on.

jufantozzi (approved):
🌳

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,645 @@
...
+		return nil, fmt.Errorf("unsupported database_type: %s" + dbtype)
+	}
+	db, err = dialectvar.connect(connectionString)
+	if err != nil {
+		return nil, fmt.Errorf("error connecting to: %s", connectionString)
+	}
+	return db, nil
+}
+
+func CreateAllTablesinDB(db *gorm.DB) (err error) {
```
mchurichi, line 40: since you are explicitly returning err, we can remove the named return values as it is redundant and confusing
```suggestion
func CreateAllTablesinDB(db *gorm.DB) error {
```

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,645 @@
...
+	err = CreateOrganizationTableInDB(db)
+	if err != nil {
+		return err
+	}
```
mchurichi, line 44: nit: we could simplify all err handlings in this file
```suggestion
	if err := CreateOrganizationTableInDB(db); err != nil;  {
		return err
	}
```

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,645 @@
...
+func CreateTrustbundleTableInDB(db *gorm.DB) error {
+	// Creates the Table for the Trustbundle Model
+	err := db.AutoMigrate(&TrustBundle{})
+	if err != nil {
+		return fmt.Errorf("sqlstore error: automigrate: %v", err)
+	}
+	return nil
+}
+
+func CreateOrganization(db *gorm.DB, org Organization) error {
```
mchurichi, line 127: is there a reason to not return the newly created organization? (the same applies to all create and update functions)
not having the org id, for instance, may be problematic when creating entities and their subentities like bridges and members

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,645 @@
...
+	}
+	br.OrganizationID = org.ID // Fill in the OrgID for the bridge
+	err = db.Where(&br).FirstOrCreate(&br).Error
+	if err != nil {
+		return fmt.Errorf("sqlstore error: %v", err)
+	}
+	return nil
+}
+
+func CreateMember(db *gorm.DB, mem Member, description string) error {
```
mchurichi, line 152: description is not being used

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,645 @@
+package sqlstore
```
mchurichi, line 1: as a general comment on this package, I don't think we should be relying on weak properties like names and descriptions to fetch entities, e.g. a function to create a membership should be passed in immutable properties like both member ids and the bridge id, not names or descriptions, that are likely to be changed.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,645 @@
...
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return nil, fmt.Errorf("sqlstore error: bridge %d does not exist in db", bridgeID)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("sqlstore error: %v. bridge query failed with bridge ID %d", err, bridgeID)
+	}
+	return &br.Memberships, nil
+}
+
+func RetrieveAllMembersbyBridgeID(db *gorm.DB, bridgeID uint) (mem *[]Member, err error) {
```
mchurichi, line 271: I think in general we shouldn't be creating a function per property we have in each entity, as this is hard to maintain. I'd propose 2 functions per entity, a simple one to fetch by id, and another that receives an object with the properties we want to fetch, something like
```go
func RetrieveMemberById(id int) (Member, error) ...
func RetrieveMembers(properties Member) ([]*Member, error) ...
```
what do you think?

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,645 @@
...
+		return nil, fmt.Errorf("sqlstore error: Member with Description=%v does not exist in DB", description)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("sqlstore error: %v", err)
+	}
+	return member, nil
+}
+
+func RetrieveMembershipbyCreationDate(db *gorm.DB, date time.Time) (*Membership, error) {
+	// RetrieveMembershipbyToken retrieves a Membership from the Database bigger than an specific date. returns an error if something goes wrong.
```
mchurichi, line 347: function comments must of above the function name

jufantozzi (approved):

mchurichi (dismissed):

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,645 @@
...
+		return err
+	}
+	err = CreateTrustbundleTableInDB(db)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func CreateOrganizationTableInDB(db *gorm.DB) error {
```
mchurichi, line 68: we agree these functions need to be part of the struct, right? probably not for this PR, but we need these grouped and behind a well-defined interface

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,266 @@
...
+		return fmt.Errorf("sqlstore error: %v", err)
+	}
+	return nil
+}
+
+func CreateBridge(db *gorm.DB, br Bridge, org Organization) error {
+	// Creates a new Bridge or ATB
+	result := db.Where("Name = ?", org.Name).First(&org) //Search if the org exists
+	if errors.Is(result.Error, gorm.ErrRecordNotFound) { // If does not, throw an error
+		return fmt.Errorf("Organization " + org.Name + " does not exist in DB")
```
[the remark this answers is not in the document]
thisab: I think this is outdated. Not present in the most recent code.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,266 @@
...
+		return fmt.Errorf("sqlstore error: %v", err)
+	}
+	return nil
+}
+
+func CreateMember(db *gorm.DB, mem Member, br Bridge) error {
+	// Creates a new Member
+	result := db.Where("Description = ?", br.Description).First(&br) //Search if the bridge exists
+	if errors.Is(result.Error, gorm.ErrRecordNotFound) {             // If does not, throw an error
+		return fmt.Errorf("Bridge Description=" + br.Description + " does not exist in DB")
```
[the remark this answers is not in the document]
thisab: I think this is outdated. Not present in the most recent code.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,266 @@
...
+		return fmt.Errorf("sqlstore error: %v", err)
+	}
+	return nil
+}
+
+func CreateMembership(db *gorm.DB, memb Membership, mem Member) error {
+	// Creates a new Membership
+	result := db.Where("SpiffeID = ?", mem.SpiffeID).First(&mem) //Search if the bridge exists
+	if errors.Is(result.Error, gorm.ErrRecordNotFound) {         // If does not, throw an error
+		return fmt.Errorf("Member SpiffeID=" + mem.SpiffeID + " does not exist in DB")
```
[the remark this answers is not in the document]
thisab: I think this is outdated. Not present in the most recent code.

pkg/server/datastore/sqlstore/sqlstore.go
thisab: Could be rewritten to return always the entity created. Looking into the Spire code, it does that. So, I think we could do that here too.

pkg/server/datastore/sqlstore/sqlstore.go
thisab: Agreed that, although unique, they could change with time.

pkg/server/datastore/sqlstore/sqlstore.go
thisab: Agreed. I rewrote to focus more on ID values.

pkg/server/datastore/sqlstore/sqlstore.go
thisab: Yes. Later on this could be part of a Plugin struct, just like Spire is.

thisab (commented):
Made changes according to mchurichi inputs.

jufantozzi:
One observation for the future: let's try to avoid large PRs as it's usually easier to deal with smaller ones 😃

thisab (commented):
Applied the requested changes to the code

thisab:
> One observation for the future: let's try to avoid large PRs as it's usually easier to deal with smaller ones smiley

Sure, we will work on that

pkg/server/datastore/sqlstore/models.go
```diff
@@ -0,0 +1,69 @@
...
+
+type Bridge struct {
+	Model
+	OrganizationID     uint
+	Description        string `gorm:"uniqueIndex"`
+	Status             bool
+	NestedFedIndicator bool
+	Memberships        []Membership `gorm:"constraint:OnDelete:CASCADE;"`
+}
+type Member struct {
```
maxlambrecht, line 28: IMO we should only add the fields we know we'll need for a POC or maybe a MVP, and then incrementally add the extra fields when the necessity for them comes, otherwise it could happen that we end up with fields in the DB that we never need and nobody knows why they are there.

thisab: Sure, I will clean it up. Then upload them again.

mchurichi:
> As discussed offline, I'll be approving this as it is to avoid making this PR even bigger, but we agreed that there's some technical debt that should be taken care of right away:
> 
>     * Basic unit tests and
> 
>     * Defining interfaces for the db plugin for CRUD operations.
> 
> 
> These interfaces will be used by the database broker (the layer above this current impl) and they are needed to properly separate the domains.

Agreed. This is not even getting merged on the main branch, and refactors and improvements can be made in smaller PRs to ease the review process.

mchurichi (approved):

PR #23, opened 2022-08-02
Initial boilerplate for telemetry

n-th:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Monitoring

**Description of change**
Add dockerfile for Prometheus usage with an initial metric server

**Which issue this PR fixes**


pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,90 @@
...
+		panic(err)
+	}
+
+	// TODO: verify telemetry plugin config to run specified plugin
+	exporter := configurePrometheusMetrics()
+	http.HandleFunc("/metrics", exporter.ServeHTTP)
+
+	go func() {
+		port := "8888"
+		c.logger.Info("Listenening on http://localhost:8888/metrics")
```
mgbcaio, line 47: nit
```suggestion
		c.logger.Info("Listening on http://localhost:8888/metrics")
```

pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,90 @@
...
+
+		address := fmt.Sprintf(":%s", port)
+		_ = http.ListenAndServe(address, nil)
```
mgbcaio, line 50: Since we do not care about the error return we could use 
```suggestion
	http.ListenAndServe(address, nil)
```

pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,90 @@
...
+	global.SetMeterProvider(exporter.MeterProvider())
+
+	return exporter
+}
+
+func FormatLabel(component, entity, action string) string {
+	return fmt.Sprintf("%s.%s.%s", component, entity, action)
+}
+
+func InitiateMeterProvider(ctx context.Context) metric.Meter {
```
mgbcaio, line 84:
```suggestion
func InitializeMeterProvider(ctx context.Context) metric.Meter {
```

mgbcaio (commented):
Good job, Nathalia! 🥇 
left a few super nit commentaries that you can agree with or not =)

pkg/harvester/manager.go
```diff
@@ -20,20 +21,27 @@ type HarvesterManager struct {
...
 func (m *HarvesterManager) Start(ctx context.Context, config config.HarvesterConfig) {
+	type key string
+
 	if m.load(config) != nil {
 		panic("unable to load configuration")
 	}
 
 	defer m.Stop()
+
+	ctxKey := key(telemetry.PackageName)
```
mchurichi, line 42: why do we need to do this?

pkg/harvester/controller/controller.go
```diff
@@ -36,16 +38,23 @@ func (c *LocalHarvesterController) Run(ctx context.Context) error {
...
+}
+
 func (c *LocalHarvesterController) run(ctx context.Context) {
 	ticker := time.NewTicker(2 * time.Second)
 
 	for {
 		select {
 		case <-ticker.C:
 			c.logger.Debug("Doing something")
+			telemetry.Count(ctx, telemetry.HarvesterController, telemetry.TrustBundle, telemetry.Add)
```
mchurichi, line 57: do we need this metric here?

conf/harvester/harvester.conf
```diff
@@ -11,3 +11,15 @@ harvester {
...
+
+# telemetry: If telemetry is desired use this section to configure the
+# available metrics collectors.
+telemetry {
+    Prometheus {
+        # host: Prometheus server host.
+        host = "localhost"
+
+        # port: Prometheus server port.
+        port = 9090
```
mchurichi, line 23: why does the harvester need to know the prometheus server host and port? isn't prometheus pulling from the harvester?

pkg/common/telemetry/config.go
```diff
@@ -0,0 +1,50 @@
...
+	Prometheus *PrometheusConfig `hcl:"Prometheus"`
+}
+
+type PrometheusConfig struct {
+	Host string `hcl:"host"`
+	Port int    `hcl:"port"`
+}
+
+// New creates a new TelemetryConfig from the given input reader.
+func New(config io.Reader) (*TelemetryConfig, error) {
```
mchurichi, line 25: not sure we need this... I think we can reuse the config parsing on harvester/config/config.go, and refer to PrometheusConfig from there, same as spire does [here](https://github.com/spiffe/spire/blob/9eb12df4d3cf2a12f92c2520d686cc705d326b7c/pkg/agent/config.go#L82) and [here](https://github.com/spiffe/spire/blob/9eb12df4d3cf2a12f92c2520d686cc705d326b7c/pkg/common/telemetry/config.go#L14)

mchurichi (commented):
Masterpiece! 💯 this looks really good Nath!

There's just one thing I think we can simplify and is the config parsing, we should be doing it in a single place.

pkg/harvester/manager.go
mchurichi: I mean, the `key` type

pkg/harvester/controller/controller.go
n-th: No, just for testing since we don't have anything to count yet.

pkg/common/telemetry/config.go
n-th: kay

pkg/harvester/manager.go
n-th: So we won't get an lint complaint when setting the ctx.WithValue. If I assign it as a string, it says "should not use built-in type string as key for value; define your own type to avoid collisions (SA1029)".

pkg/common/telemetry/names.go
```diff
@@ -0,0 +1,31 @@
+package telemetry
+
+// package name
+const (
+	Harvester = "harvester"
+	Galadriel = "harvester"
```
jufantozzi, line 6: typo?
```suggestion
	Galadriel = "galadriel-server"
```

pkg/common/telemetry/names.go
```diff
@@ -0,0 +1,31 @@
+package telemetry
+
+// package name
+const (
+	Harvester = "harvester"
```
jufantozzi, line 5:
```suggestion
	Harvester = "galadriel-harvester"
```

conf/harvester/harvester.conf
n-th: Ah, SPIRE uses a sink pattern to provide metrics to Prometheus, since we are not using this one I will remove the configuration.

jufantozzi:
Just waiting on the merge conflict to be resolved to approve 🍰

jufantozzi (dismissed):
🐰

jufantozzi (approved):

PR #24, opened 2022-08-02
[Server] - Adds `server run` command to server CLI

miguelfferraz:
Signed-off-by: Miguel Figueira Ferraz <miguelfigueiraferraz@gmail.com>

**Pull Request checklist**

- [ ] Documentation updated?

**Affected functionality**
Adds a new command `server run` to CLI, this command starts the Galadriel Server.

**Description of change**
The code inside `/cmd/server/main.go` was moved to `/cmd/server/api/api.go`. Now main.go only starts the CLI, while api.go handles starting Galadriel Server.
A new package called `cli` has been created inside `/cmd/server`. This package contains all CLI commands.


cmd/server/cli/run.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"github.com/HewlettPackard/Galadriel/cmd/server/api"
+	"github.com/spf13/cobra"
+)
+
+var runCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Start HTTP Galadriel server",
+	Long: `Run this command to start HTTP Galadriel server`,
```
mchurichi, line 11: super nit
```suggestion
	Long:  "Run this command to start HTTP Galadriel server",
```

cmd/server/cli/server.go
```diff
@@ -0,0 +1,15 @@
+package cli
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var serverCmd = &cobra.Command{
+	Use:   "server",
+	Short: "Commands to manage Galadriel server",
+	Long: `Commands to manage Galadriel server`,
```
mchurichi, line 10: super nit
```suggestion
	Long:  "Commands to manage Galadriel server",
```

mchurichi (commented):
Nicely done, Miguel! 🙌 

It seems that there are a few files that are missing a gofmt pass, could you give it a run? Other than that, this looks great to me.

mchurichi:
And one more thing, could you add a super simple readme for the Server CLI? the idea is to document it from the very beginning, something like the [Harvester doc](https://github.com/HewlettPackard/galadriel/blob/master/doc/harvester.md)

mchurichi:
@mgbcaio any thoughts here?

n-th:
@miguelfferraz , can you add a help command as well as unit tests?

mgbcaio:
I agree with @mchurichi and @n-th. Other than that, I think we could follow the same development pattern @mchurichi implemented regarding the harvester CLI -> [Maxi PR](https://github.com/HewlettPackard/galadriel/pull/16/files). What do you guys think?

mchurichi:
Just to be clear, the CLI pushed in the [Boilerplate PR](https://github.com/HewlettPackard/galadriel/pull/16/files) was only intended to define the basic directories and interfaces architecture. The tooling choices (no external lib, in the case of the CLI) shouldn't be taken into account.

That said, I like the idea of using a specialized CLI library like cobra or viper for this, but I'll rely on @miguelfferraz and @mgbcaio's (who's working on the CLI of the harvester) proposals for that, you guys should agree on tooling and style to have the same stuff on both Server and Harvester side.

mgbcaio:
> Just to be clear, the CLI pushed in the [Boilerplate PR](https://github.com/HewlettPackard/galadriel/pull/16/files) was only intended to define the basic directories and interfaces architecture. The tooling choices (no external lib, in the case of the CLI) shouldn't be taken into account.
> 
> That said, I like the idea of using a specialized CLI library like cobra or viper for this, but I'll rely on @miguelfferraz and @mgbcaio's (who's working on the CLI of the harvester) proposals for that, you guys should agree on tooling and style to have the same stuff on both Server and Harvester side.

That makes sense, thanks for clearing that out, Maxi! Considering that the Harverter CLI will follow the Cobra library, just like @miguelfferraz its doing.

cmd/server/cli/root.go
```diff
@@ -0,0 +1,19 @@
...
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
```
mchurichi, line 17: what do you think about moving this to cmd/server/main.go? the CLI should be environment-agnostic... we can have this function returning an integer, and the main fn is the one that spits the return code back to the shell. This will make facilitate CLI testing without having to mock the os.Exit call.

miguelfferraz: > what do you think about moving this to cmd/server/main.go? the CLI should be environment-agnostic... we can have this function returning an integer, and the main fn is the one that spits the return code back to the shell. This will facilitate CLI testing without having to mock the os.Exit call.

Thanks for this suggestion, that makes sense. I moved os.Exit call to cmd/server/main.go.

miguelfferraz:
> @miguelfferraz , can you add a help command as well as unit tests?

@n-th Cobra already generates a help command and a --help flag for all commands in the CLI.
I'm working on the unit tests.

cmd/server/cli/root.go
```diff
@@ -0,0 +1,20 @@
+package cli
+
+import (
```
jufantozzi, line 3: I think it's worth to already start splitting each one of these into their own packages to avoid big refactors afterwards

cmd/server/cli/run.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/HewlettPackard/Galadriel/cmd/server/api"
+	"github.com/spf13/cobra"
+)
+
+var runServerFn = api.Run
+
+func NewRunCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "run",
+		Short: "Start HTTP Galadriel server",
```
jufantozzi, line 13:
```suggestion
		Short: "Runs the Galadriel server",
```

cmd/server/cli/run.go
```diff
@@ -0,0 +1,23 @@
...
+		Long:  "Run this command to start HTTP Galadriel server",
```
jufantozzi, line 14:
```suggestion
		Long:  "Run this command to start the Galadriel server",
```

jufantozzi (changes requested):
Thank you @miguelfferraz! Nice work! this looks great. I just left some minor observations but once we're done with those we're good to go 👌

cmd/server/cli/run_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	runServerFn = func() {
+		called = true
+	}
+
+	cmd := NewRunCmd()
+	err := cmd.Execute()
+
+	if err != nil {
+		fmt.Printf("%s", err)
```
mchurichi, line 21: we should be using an assert here, and making sure that if the function call returns an error the test must fail

jufantozzi:
Hey miguel, I tried building and running but the command prompt doesn't seem to return anything for any command. I didn't have enough time to debug it yet, PTAL.

cmd/server/api/api.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/labstack/echo/v4"
+	echomiddleware "github.com/labstack/echo/v4/middleware"
+)
+
+func Run() {
+
+	var port = flag.Int("port", 33208, "Port for HTTP Galadriel server")
+	flag.Parse()
+
+	// Clear out the servers array in the swagger spec, that skips validating
```
maxlambrecht, line 16: Remove all the unnecessary comments and commented code.

miguelfferraz:
> Hey miguel, I tried building and running but the command prompt doesn't seem to return anything for any command. I didn't have enough time to debug it yet, PTAL.

Thanks for the warning Juliano, the CLI is really with a bug on execution.

jufantozzi (changes requested):
Threw a few more nits, but the only real concern here is allowing the server to start without properly loading the config.

> The problem is because I split the commands into different packages, I didn't figure out how to get this work.

AFAIK the way to go here would be to manually add the commands to the rootCmd, but it does look less clean IMO... We shouldn't have a problem going with the way it is, at least for now.

jufantozzi:
Oh, and please @miguelfferraz  make sure to have resolve the conflicts and to have signed all commits so we don't get dismissed approves.

jufantozzi:
Hi Miguel, we're almost good to go! Just need to sign the commits now. 

If you're unsure about what I'm referring to, [here's a reference](https://superuser.com/a/1123928) (this assumes you have gpg signing set up, the `-S` flag)

conf/server/server.conf
```diff
@@ -0,0 +1,13 @@
+server {
+    # spire_socket_path: SPIRE Server Socket of the instance to manage.
+    # Default: /tmp/spire-server/private/api.sock
+    spire_socket_path = "/tmp/spire-server/private/api.sock"
```
mchurichi, line 4: please remove this setting, we won't need to connect to the spire server from the galadriel server

conf/server/server.conf
```diff
@@ -0,0 +1,13 @@
...
+
+    # server_address: Upstream Galadriel Server DNS name or IP address with port.
+    # E.g: localhost:8080, my-upstream-server.com:4556, 192.168.1.125:4000
+    server_address = "localhost:8080"
```
mchurichi, line 8:
```suggestion
    # listen_address: DNS name or IP address with port for the Galadriel Server to listen on
    # E.g: localhost:8080, my-upstream-server.com:4556, 192.168.1.125:4000
    # Default: localhost:8080
    listen_address = "localhost:8080"
```

pkg/server/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/hashicorp/hcl"
+	"github.com/pkg/errors"
+)
+
+type Server struct {
+	ServerConfigSection *ServerConfigSection `hcl:"server"`
+}
+
+type ServerConfigSection struct {
+	SpireSocketPath string `hcl:"spire_socket_path"`
```
mchurichi, line 16: remove

pkg/server/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+	ServerAddress   string `hcl:"server_address"`
```
mchurichi, line 17:
```suggestion
	ListenAddress   string `hcl:"listen_address"`
```

pkg/server/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return nil
+}
+
+func (c *Server) setDefaults() {
+	if c.ServerConfigSection.LogLevel == "" {
+		c.ServerConfigSection.LogLevel = "INFO"
+	}
+
+	if c.ServerConfigSection.SpireSocketPath == "" {
```
mchurichi, line 68: remove

pkg/server/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+		c.ServerConfigSection.SpireSocketPath = "/tmp/spire-server/private/api.sock"
+	}
```
mchurichi, line 70: please add ListenAddress default value to `localhost:8080`

pkg/server/config/disk.go
```diff
@@ -0,0 +1,19 @@
+package config
```
mchurichi, line 1: not for this PR, but we need to move this to a common package to be shared with the Harvester and Server to avoid code duplication

mchurichi:
Hi @miguelfferraz, I'm very sorry for the delay in getting this merged in 🙏

Just left some comments about the config file. There are a few things we won't need on the Server side.
I think we'll be good to merge this once we get those changes addressed.

Happy to discuss this offline on Slack or in a call if needed. Feel free to reach out.

pkg/server/config/disk.go
jufantozzi: ...and to properly separate domains and avoid coupling. This should be addressed in other PR's soon.

cmd/server/cli/run.go
```diff
@@ -0,0 +1,51 @@
...
+		Use:   "run",
+		Short: "Runs the Galadriel server",
+		Long:  "Run this command to start the Galadriel server",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			configPath, err := cmd.Flags().GetString("config")
+			if err != nil {
+				return err
+			}
+
+			runServerFn(configPath)
```
mchurichi, line 27: one last request: can we have the same return pattern as in the harvester? let's try to keep both harvester and server code as close as possible
https://github.com/HewlettPackard/galadriel/pull/37/files#diff-c89d406f570ba5152de10698eb3b4ed24702ae7d03eb6c95563e832bffe2d638R26

mchurichi: and I swear this is the last one 😆

mchurichi (approved):
Excellent work @miguelfferraz! 🏅 Thank you so much! 🙏

jufantozzi (approved):
Thank you so much for your hard work and patience @miguelfferraz!

PR #26, opened 2022-08-08
Spire Server broker implementation

mchurichi:
Defines SPIRE Server interface and implements SPIRE Server broker.

This is still WIP, but I would love to have your early review.
Still missing unit tests.


jufantozzi (commented):
Amazing work so far @mchurichi! 🌟

maxlambrecht:
I know it's not the main point of this PR, but I think we should define which CLI framework we are going to use, if any, and stick to that option in all our PRs with CLI code. I'd suggest using [Cobra](https://cobra.dev/), it's the one used by Kubernetes CLIs and Docker. Another option is [github.com/mitchellh/cli](https://github.com/mitchellh/cli), used by Hashicorp tools and SPIRE, but it has a more complex API.

mchurichi:
> I know it's not the main point of this PR, but I think we should define which CLI framework we are going to use, if any, and stick to that option in all our PRs with CLI code. I'd suggest using [Cobra](https://cobra.dev/), it's the one used by Kubernetes CLIs and Docker. Another option is [github.com/mitchellh/cli](https://github.com/mitchellh/cli), used by Hashicorp tools and SPIRE, but it has a more complex API.

Agreed. #24 is implementing the CLI using Cobra for the first CLI command. The implementation of this PR is mostly to ease testing, but I'd rather remove all CLI code before marking this PR as ready and have the Harvester-SPIRE CLI commands implemented on a subsequent PR.

maxlambrecht:
> I'd rather remove all CLI code before marking this PR as ready and have the Harvester-SPIRE CLI commands implemented on a subsequent PR.

It makes sense to me.

jufantozzi:
@maxlambrecht thanks for this clarifications. My take on this is that, _ideally_, we could consider using [github.com/mitchellh/cli](https://github.com/mitchellh/cli) to strive for compatibility with SPIRE, while using Cobra could cause friction. This is, though, really the only argument I can think of for supporting using it instead of Cobra.

maxlambrecht:
> @maxlambrecht thanks for these clarifications. My take on this is that, _ideally_, we could consider using [github.com/mitchellh/cli](https://github.com/mitchellh/cli) to strive for compatibility with SPIRE, while using Cobra could cause friction. This is, though, really the only argument I can think of for supporting using it instead of Cobra.

The CLI framework is an implementation decision that doesn't affect compatibility with SPIRE since the integration with SPIRE is through public APIs (REST, gRPC, etc). I'd go with Cobra as it seems to me that it's less complex and it's also use by well established and big open source projects like Kubernetes and Docker.

jufantozzi:
> The CLI framework is an implementation decision that doesn't affect compatibility with SPIRE since the integration with SPIRE is through public APIs (REST, gRPC, etc)

Right, "compatibility" was a terrible choice for word there, and I wasn't thinking of compatibility in that sense. I mean our code architecture, frameworks that we use and other decisions that we take, could make harder to incorporate Galadriel into SPIRE, if that ever happen. This is of course something we don't know if it will happen, neither if that makes sense, but  just wanted to lay out these thoughts. But yes, I also agree on Cobra.

maxlambrecht:
> > The CLI framework is an implementation decision that doesn't affect compatibility with SPIRE since the integration with SPIRE is through public APIs (REST, gRPC, etc)
> 
> Right, "compatibility" was a terrible choice for word there, and I wasn't thinking of compatibility in that sense. I mean our code architecture, frameworks that we use and other decisions that we take, could make harder to incorporate Galadriel into SPIRE, if that ever happen. This is of course something we don't know if it will happen, neither if that makes sense, but just wanted to lay out these thoughts. But yes, I also agree on Cobra.

I don't foresee Galadriel code being incorporated into SPIRE core, by the principle of separation of duties it wouldn't make sense to add that amount complexity to SPIRE; the same happened with the `spire-controller-manager` used to generate the registration entries in SPIRE, it's not part of the SPIRE code, it's a separate binary and the code is in a separate repo.

mchurichi:
Closing this in favor of #41 and #42

PR #30, opened 2022-08-11
[Server] - API interface - Implementation

thisab:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**
Defined the Server and made the Server implement (dummy initially) functions for the API

**Which issue this PR fixes**

### DIFF
--- a/pkg/server/api/management/handler.go
+++ b/pkg/server/api/management/handler.go
@@ -4,129 +4,129 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
-type Server struct {
+type Management struct {
 }
 
-func NewServer() *Server {
-	return &Server{}
+func NewManagement() *Management {
+	return &Management{}
 }
 
 // (GET /spireServers)
-func (server Server) GetSpireServers(ctx echo.Context, params GetSpireServersParams) error {
+func (m *Management) GetSpireServers(ctx echo.Context, params GetSpireServersParams) error {
 	return nil
 }
 
 // (POST /spireServers)
-func (server Server) CreateSpireServer(ctx echo.Context) error {
+func (m *Management) CreateSpireServer(ctx echo.Context) error {
 	return nil
 }
 
 // (DELETE /spireServers/{spireServerId})
-func (server Server) DeleteSpireServer(ctx echo.Context, spireServerId int64) error {
+func (m *Management) DeleteSpireServer(ctx echo.Context, spireServerId int64) error {
 	return nil
 }
 
 // (PUT /spireServers/{spireServerId})
-func (server Server) UpdateSpireServer(ctx echo.Context, spireServerId int64) error {
+func (m *Management) UpdateSpireServer(ctx echo.Context, spireServerId int64) error {
 	return nil
 }
 
 // (PUT /trustBundles/{trustBundleId})
-func (server Server) UpdateTrustBundle(ctx echo.Context, trustBundleId int64) error {
+func (m *Management) UpdateTrustBundle(ctx echo.Context, trustBundleId int64) error {
 	return nil
 }
 
 // GetFederationGroupMembershipsParams defines parameters for GetFederationGroupMemberships.
-func (server Server) GetFederationGroupMemberships(ctx echo.Context, params GetFederationGroupMembershipsParams) error {
+func (m *Management) GetFederationGroupMemberships(ctx echo.Context, params GetFederationGroupMembershipsParams) error {
 	return nil
 }
 
 // (POST /federationGroupMemberships)
-func (server Server) CreateFederationGroupMembership(ctx echo.Context) error {
+func (m *Management) CreateFederationGroupMembership(ctx echo.Context) error {
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/server/api/management/handler.go
```diff
@@ -1,35 +1,132 @@
 package management
 
 import (
-	"errors"
-
 	"github.com/labstack/echo/v4"
 )
 
-type MyDumbServer struct {
+type Server struct {
```
mchurichi, line 7: can we define the actual interface this struct implements?

pkg/server/api/management/handler.go
```diff
@@ -1,35 +1,132 @@
...
+	return nil
+}
+
+// (GET /organizations/{orgID})
+func (server Server) GetOrgbyID(ctx echo.Context, orgID int64) error {
+	return nil
+}
+
+// (PUT /organizations/{orgID})
+func (server Server) UpdateOrganizaion(ctx echo.Context, orgID int64) error {
```
n-th, line 130: UpdateOrganization

pkg/server/api/management/handler.go
```diff
@@ -1,35 +1,132 @@
...
+	return nil
+}
+
+// (GET /federationRelationships/{relationshipID})
+func (server Server) GetFederationRelationshipbyID(ctx echo.Context, relationshipID int64) error {
+	return nil
+}
+
+// (PUT /federationRelationships/{relationshipID})
+func (server Server) UpdateFederationRelationshipship(ctx echo.Context, relationshipID int64) error {
```
n-th, line 105: UpdateFederationRelationship

pkg/server/api/management/handler.go
```diff
@@ -1,35 +1,132 @@
...
+	return nil
+}
+
+// (POST /federationGroupMemberships)
+func (server Server) CreateFederationGroupMembership(ctx echo.Context) error {
+	return nil
+}
+
+// (DELETE /federationGroupMemberships/{membershipID})
+func (server Server) DeletefederationGroupMembership(ctx echo.Context, membershipID int64) error {
```
n-th, line 50: DeleteFederationGroupMembership

n-th: This seems to happen to all federationGroupMembership related functions

n-th (changes requested):
I found some typos in the function, could you revise the .yamls that generate those functions for misspells?

pkg/server/api/management/handler.go
thisab: Indeed. To do that, we would need to include management.yaml file and correct it also.

pkg/server/api/management/handler.go
```diff
@@ -1,35 +1,132 @@
...
+}
+
+func NewServer() *Server {
+	return &Server{}
 }
 
 // (GET /spireServers)
-func (server MyDumbServer) GetSpireServers(ctx echo.Context, params GetSpireServersParams) error {
-	return errors.New("Ah, não, deu erro")
+func (server Server) GetSpireServers(ctx echo.Context, params GetSpireServersParams) error {
```
maxlambrecht, line 15: What's the reason for not using the `context.Context` from the standard library?

thisab: @maxlambrecht we are using context of the Echo library, following just like the main.go does nowadays.
Do you think we should not follow that ?

jufantozzi: AFAIK this is how the lib works, this `ctx` will have the request details

maxlambrecht: We should use the `context.Context` from the standard library in all our APIs definitions. Our APIs should be as standard as possible, if in our implementation we use some library that needs a specific context struct, we'll need to "adapt" the standard ctx to it, but that specific context is an implementation detail that must not be leaked into the public interface.

wiliamsouza: I was doing a little research to discover a way to keep using the echo as our web framework but it seems like there is no way to use it without using it’s  own context (`echo.Context` implementation[1]). Searching around I have found out that what other people do is to adapt/convert from `echo.Context` to Go lang `context.Context` before calling downstream code. This way it decouples the inner layers from your web framework API code. The main benefits of this approach is to isolate/decouple web frameworks or ORM from Galadriel  main business logic. May I upload the suggested change to this PR showing that solution?

PS: Maybe use the Go lang `internal` package to enforce that isolation[3] can be an option.

[1] https://stackoverflow.com/questions/47843914/transforming-echo-context-to-context-context
[2] https://github.com/labstack/echo/issues/1175#issuecomment-455226401
[3] https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface

maxlambrecht: You are right, it seems that we need to use the `echo.Context`.

thisab:
@jufantozzi I think this PR is a bit outdated. Does it still make sense to leave it open?

jufantozzi:
@thisab you're right, we can close this

PR #39, opened 2022-08-26
fix go-check target

jufantozzi:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

fixed make target `go-check`;
renamed `new` function to avoid collision with builtin `new` and to keep consistency with server. 

**Description of change**

**Which issue this PR fixes**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,8 @@ oapi_codegen_dir = $(build_dir)/protoc/$(protoc_version):q
 
 go-check:
 ifeq (go$(go_version), $(shell $(go_path) go version 2>/dev/null | cut -f3 -d' '))
-	@echo "Installing go$(go_version)..."
+else
+	@echo "Installing go $(go_version)..."
 	$(E)rm -rf $(dir $(go_dir))
 	$(E)mkdir -p $(go_dir)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -75,7 +75,8 @@ oapi_codegen_dir = $(build_dir)/protoc/$(protoc_version):q
 
 go-check:
 ifeq (go$(go_version), $(shell $(go_path) go version 2>/dev/null | cut -f3 -d' '))
-	@echo "Installing go$(go_version)..."
```
mchurichi, line 78: sorry but this is simplified Chinese to me 😆 could you explain what is this doing and how this change impacts the target?

jufantozzi: Running `make build` today will trigger this target every time, instead of doing what it actually should

n-th (approved):

maxlambrecht (approved):

maxlambrecht:
Peeking into the Makefile it seems there are several things that are not used, and probably won't be used, that just add extra complexity, e.g, windows related stuff. I suggest removing all unnecessary stuff from Makefile.

PR #40, opened 2022-08-29
[Server] - Definition of the DB structure

thisab:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Does not affect any other component at the moment.
There are two main definitions that this code also depends:
[1] - The use of context. Here the use of the default context library is assumed.
[2] - The API spec. The following mappings are assumed to the DB: SpireServer = Member, FederationGroupMembership = Membership, FederationRelationship = Relationship.

**Description of change**
This PR intends to define the structure of the DB code.
For example:
datastore.go defines the interface to the DB based on the API spec and model definition.
sqlstore.go would hold the implementation of the SQL code


pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,55 @@
...
+	management "github.com/HewlettPackard/galadriel/pkg/server/api/management"
+	"github.com/labstack/echo/v4"
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/spiffe/spire/proto/spire/common"
+)
+
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(echo.Context, *management.SpireServer) (*management.SpireServer, error)
+	CreateMembership(ctx context.Context, membership *management.FederationGroupMembership, memberID uint, bridgeID uint) (*management.FederationGroupMembership, error)
```
jufantozzi, line 17: an example would be:
```suggestion
	CreateMembership(context.Context, membership *management.FederationGroupMembership, MemberID, BridgeID) (*management.FederationGroupMembership, error)
```
and have defined
`type MemberID uint`
`type BridgeID uint`

jufantozzi (commented):
Hey @thisab! great work, I think we're on the right direction here. I left some nits for now.

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,55 @@
...
+
+	management "github.com/HewlettPackard/galadriel/pkg/server/api/management"
+	"github.com/labstack/echo/v4"
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/spiffe/spire/proto/spire/common"
+)
+
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(echo.Context, *management.SpireServer) (*management.SpireServer, error)
```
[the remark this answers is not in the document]
thisab: The style is similar to the code in [spire](https://github.com/spiffe/spire/blob/main/pkg/server/datastore/datastore.go) . I agree that, for example, FederationRelationship could be defined in the spec. 
Whenever the spec is updated we could also update this code.

For the echo, yes that should not be there. At least not mixing contexts between echo and standard. Fixing that.

pkg/server/datastore/datastore.go
jufantozzi: As discussed offline with @thisab, this can be better looked at once we have an updated version of the API specs and definitions. Let's try to have all these types defined in there.

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,55 @@
...
+	CreateMembership(ctx context.Context, membership *management.FederationGroupMembership, memberID uint, bridgeID uint) (*management.FederationGroupMembership, error)
+	CreateTrustBundle(ctx context.Context, trust *common.Bundle, memberID uint) (*common.Bundle, error)
+	CreateRelationship(ctx context.Context, newrelation *FederationRelationship, sourceID uint, targetID uint) error
+	RetrieveBridgebyID(ctx context.Context, brID uint) (*management.FederationGroup, error)
+	RetrieveAllMembershipsbyBridgeID(ctx context.Context, bridgeID uint) (*[]management.FederationGroupMembership, error)
+	RetrieveAllMembersbyBridgeID(ctx context.Context, bridgeID uint) (*[]management.SpireServer, error)
+	RetrieveAllMembershipsbyMemberID(ctx context.Context, memberID uint) (*[]management.FederationGroupMembership, error)
```
mchurichi, line 23: how about a single `RetrieveMembers(ctx context.Context, params RetrieveMembersParam)`?
this can serve many purposes with a single implementation.

the same applies to all others `[Retrieve|Delete]XXXXbyNNN` methods

jufantozzi:
One NIT that I'd like to propose that we standardize on a naming format:
`func FooBarByID(...)` with capital By and ID

pkg/server/datastore/sqlstore/model.go
```diff
@@ -0,0 +1,47 @@
...
+	CreatedAt time.Time
+	UpdatedAt time.Time
+}
+
+type Member struct {
+	Model
+	Description   string `gorm:"uniqueIndex"`
+	Memberships   []Membership
+	Relationships []Relationship `gorm:"constraint:OnDelete:CASCADE;"`
+	TrustBundles  []TrustBundle  `gorm:"constraint:OnDelete:CASCADE;"`
```
mchurichi, line 18: don't we need a non-unique field for the trust domain here?

pkg/server/datastore/sqlstore/model.go
```diff
@@ -0,0 +1,47 @@
...
+	TargetMemberID   uint
+	Status           string
+	RelationshipType string
+	TTL              uint
+}
+
+type TrustBundle struct {
+	Model
+	MemberID    uint //Implicit Foreign Key
+	TrustBundle string
```
mchurichi, line 40: some open questions: how do we know which is the _current_ bundle of a given member? do we keep the old ones as they rotate?

pkg/server/datastore/datastore.go
thisab: One thing that I noticed there is functions with Bridge entity, which I think we could remove for now (just for the scope of the PoC).

If I understood the idea: create a new struct called RetrieveMembersParam?
In line with this idea, we could pass a Member entity as a parameter to a function (say RetrieveAllMembershipsByID) and the argument and return struct themselves are self-descriptive: Retrieve all Memberships of a FederationGroup that has a particular MemberID.

Or.. if we stick to using memberID, like some of the [Spire code does with some variables like SpiffeID](https://github.com/spiffe/spire/blob/main/pkg/server/datastore/datastore.go), we could rename the function above to RetrieveAllMembershipsByID , as the Member is already explicit in the argument.

mchurichi: > If I understood the idea: create a new struct called RetrieveMembersParam? In line with this idea, we could instead pass a Member entity as a parameter to a function (say RetrieveAllMembershipsByID) and the argument and return struct themselves are self-descriptive: Retrieve all Memberships of a FederationGroup that has a particular MemberID.

You got the idea. The suggestion was to create a new struct `RetrieveMembersParam` that holds all the parameters you can filter by. This way we gain flexibility by having a single method signature we can grow as needed (adding a new filter will be adding a new field on the struct and the logic to handle it).
Think of it as query params on a REST call: you don't usually add a new path when adding a new filter on a given entity, you just add code to take care of the new field being passed via query.

See how [spire](https://github.com/spiffe/spire/blob/e8b280bbda295f45d68b683b679edf3e13cda445/pkg/server/datastore/datastore.go#L32) implements the `ListXXX` methods.

> Or.. if we stick to using memberID, like some of the [Spire code does with some variables like SpiffeID](https://github.com/spiffe/spire/blob/main/pkg/server/datastore/datastore.go), we could rename the function above to RetrieveAllMembershipsByID , as the Member is already explicit in the argument.

I'd rather keep it as it is, as when reading `byID` sounds like you're filtering the Membership ID.

tldr; if we have multiple methods receiving different params but getting the same output we could have a single method and use a struct to hold the params.

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/spiffe/spire/proto/spire/common"
+)
+
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(echo.Context, *management.SpireServer) (*management.SpireServer, error)
+	CreateMembership(ctx context.Context, membership *management.FederationGroupMembership, memberID uint, bridgeID uint) (*management.FederationGroupMembership, error)
+	CreateTrustBundle(ctx context.Context, trust *common.Bundle, memberID uint) (*common.Bundle, error)
+	CreateRelationship(ctx context.Context, newrelation *FederationRelationship, sourceID uint, targetID uint) error
```
mchurichi, line 19: I believe we should be consistent with the returns of the Create and Update methods. There are a number of reasons to return the object being created/updated as a result of the operation, like db/backend populated fields (create datetime, autogenerated ids, calculated fields) and easiness to use from the client side.

pkg/server/datastore/sqlstore/model.go
jufantozzi: IMO this hold the latest bundle. My take is that might be worth to keep old ones, but making sure we're not stacking them forever, probably hold the latest X?

pkg/server/datastore/datastore.go
thisab: Yes, agreed. The Create/Update method should also return the object

pkg/server/datastore/sqlstore/model.go
thisab: Or maybe the trustdomain is inside the TrustBundle?

mchurichi: The Trust Domain is defined in the Bundle itself, but it's inherent to the Spire server(s) (the Member) and it doesn't change on bundle rotation (if it does means there's something wrong and we should evict the Harvester and request re-enrollment)

pkg/server/datastore/sqlstore/model.go
mchurichi: Being defensive I'd say we should keep X old ones and run periodic cleanup jobs as you suggested, but don't really see a use case for that (disaster recovery? rollback?).
If we only need the latest one we could have it as a field of the Member.

An additional thought: do we need to have fields for certificate issuance, expiration dates, and/or CA TTL? that way we can proactively take action on bundles close to expiration without the need of parsing the bundles each time.

@maxlambrecht any thoughts on this?

pkg/server/datastore/sqlstore/model.go
thisab: Oh, ok, I think this is related to the discussion below then.

mchurichi: Mmm not sure they are strictly related topics, just to be clear when I say Trust Domain I mean the string `example.org` , `production.com`, etc, not the whole bundle.
As an example, we'll likely have an endpoint to fetch the current bundle or the federation relationships of a member given its trust domain. We'd need this TD handy for querying purposes.

thisab: What you mean is related to [line 195](https://github.com/spiffe/spire/blob/main/pkg/server/datastore/datastore.go) then?

pkg/server/datastore/datastore.go
thisab: I think I understood what you meant. On Spire I think there are many functions that has many selectors
For now I believe we would only use ID as a selector. What do you think if we mantain it with "ByID" for now and refactor it later on when we have many other selectors?

mchurichi: when you say to keep the `ByID`s you mean only `RetrieveXXXByID` or are you talking about `RetrieveXXXByYYYID` too?

thisab: @maxlambrecht Well, we could leave RetrieveXXXByID by default as the argument implies the YYY, right?

mchurichi: I'm asking if we want to keep both eg `RetrieveMembershipById(membershipID uint)` AND `RetrieveAllMembershipsByMemberID(memberID uint)`. They are not the same thing.

Just to make sure we are on the same page, my suggestion was to replace this:

```go
RetrieveMembershipByID(ctx context.Context, id uint) (*management.FederationGroupMembership, error)
RetrieveMembershipByCreationDate(ctx context.Context, date time.Time) (*management.FederationGroupMembership, error)
RetrieveMembershipByToken(ctx context.Context, token string) (*management.FederationGroupMembership, error)
```

with this:

```go
RetrieveMembershipByID(ctx context.Context, id uint) (*management.FederationGroupMembership, error) // this one remains the same
RetrieveMemberships(ctx context.Context, params RetrieveMembershipsParams) ([]*management.FederationGroupMembership, error)

type RetrieveMembershipsParams struct {
	AfterCreationDate  time.Time
	BeforeCreationData time.Time
	ByToken 	   string
	ByMemberID         uint
	// etc
}
```

thisab: Yes, I understood that too. 
What Im proposing is to maintain the ByID for all functions, like:
RetrieveMembershipByID(ctx context.Context, id uint) (*management.FederationGroupMembership, error)
RetrieveMembershipsByID(ctx context.Context, id uint) ([]*management.FederationGroupMembership, error)

and refactor it later on when (of if) we use many other selectors?

thisab: Ive pushed a change with the suggestion for evaluation

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,48 @@
...
+
+	if ds.db, err = dialectvar.connect(connectionString); err != nil {
+		return fmt.Errorf("error connecting to: %s", connectionString)
+	}
+	return migrateDB(ds.db)
+}
+
+// Implements the CreateMember function from Datastore
+// Creates a new Member in the database. Returns error if fails.
+func (ds *Plugin) CreateMember(ctx echo.Context, server *management.SpireServer) (*management.SpireServer, error) {
```
mchurichi, line 39: this should be implementing the DataStore interface, right? if so, the arguments and return types should match the CreateMember interface: `CreateMember(context.Context, *entity.Member) (*entity.Member, error)`
additionally we should also add the `OpenDB` function to the interface

pkg/common/entity/federationgroups.go
```diff
@@ -0,0 +1,8 @@
+package entity
+
+type FederationGroup struct {
+	Id     uint
+	Name   string
+	Orgid  uint
```
mchurichi, line 6: nit: `OrganizationId`

pkg/common/entity/membership.go
```diff
@@ -0,0 +1,8 @@
+package entity
+
+type Membership struct {
+	FederationGroupId uint
+	Id                uint
+	SpireServerId     uint
```
mchurichi, line 6:
```suggestion
	MemberId     uint
```

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,35 @@
...
+	entity "github.com/HewlettPackard/galadriel/pkg/common/entity"
+	"github.com/HewlettPackard/galadriel/pkg/server/api/management"
+	"github.com/spiffe/spire/proto/spire/common"
+)
+
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(context.Context, *entity.Member) (*entity.Member, error)
+	CreateMembership(ctx context.Context, membership *entity.Membership) (*entity.Membership, error)
+	CreateTrustBundle(ctx context.Context, trust *common.Bundle, memberID uint) (*common.Bundle, error)
```
mchurichi, line 15:
```suggestion
	CreateTrustBundle(ctx context.Context, trustBundle *common.Bundle, memberID uint) (*common.Bundle, error)
```

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,35 @@
...
+	CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceID uint, targetID uint) (*entity.Relationship, error)
```
mchurichi, line 16:
```suggestion
	CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceMemberId uint, targetMemberId uint) (*entity.Relationship, error)
```

mchurichi:
This is super picky but wanted to bring up a thing we were talking about with @jufantozzi just yesterday. It's about `ID` vs `Id`. There are a few forms for this:
- `Id`/`ID` as in a table primary key
- `MemberId`/`MemberID` as a reference to another table or entity's id
- `RetrieveMemberById`/`RetrieveMemberByID` as a public API of the DataStore

I don't have a strong opinion but I do believe we must choose a convention and stick to it.

Thoughts?

pkg/server/datastore/sqlstore/model.go
wiliamsouza: One suggestion is to model this tabel as 1-to-1 with some other table(I can't figure witch one) and have a ´TrustBundleHistory´ when we keep the old ´TrustBundle´ and the periodic cleanup job running again that history table.

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	entity "github.com/HewlettPackard/galadriel/pkg/common/entity"
+	"github.com/HewlettPackard/galadriel/pkg/server/api/management"
+	"github.com/spiffe/spire/proto/spire/common"
+)
+
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(context.Context, *entity.Member) (*entity.Member, error)
```
wiliamsouza, line 13: Maybe let parameters name more explicit? ´CreateMember(context.Context, member *entity.Member) (*entity.Member, error)´

pkg/server/datastore/datastore.go
wiliamsouza, line 16: Here you can use share the type for last two parameters:
`CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceMemberId , targetMemberId uint) (*entity.Relationship, error)`

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,35 @@
...
+	CreateMembership(ctx context.Context, membership *entity.Membership) (*entity.Membership, error)
+	CreateTrustBundle(ctx context.Context, trust *common.Bundle, memberID uint) (*common.Bundle, error)
+	CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceID uint, targetID uint) (*entity.Relationship, error)
+	RetrieveMembershipsByID(ctx context.Context, memberID uint) (*[]management.FederationGroupMembership, error)
+	RetrieveRelationshipsByID(ctx context.Context, memberID uint) (*[]entity.Relationship, error)
+	RetrieveTrustBundlesByID(ctx context.Context, memberID uint) (*[]common.Bundle, error)
+	RetrieveMemberByID(context.Context, entity.Member) (*entity.Member, error)
```
wiliamsouza, line 20: `RetrieveMemberByID(context.Context, member entity.Member) (*entity.Member, error)`?

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,35 @@
...
+	RetrieveMembershipByID(ctx context.Context, membershipID uint) (*management.FederationGroupMembership, error)
+	RetrieveRelationshipBySourceandTargetID(ctx context.Context, source uint, target uint) (*entity.Relationship, error)
+	RetrieveTrustbundleByMemberID(ctx context.Context, memberID string) (*common.Bundle, error)
+	UpdateMember(context.Context, management.SpireServer) error
```
wiliamsouza, line 24: `UpdateMember(context.Context, *entity.Member) error`?

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,35 @@
...
+	UpdateMembership(context.Context, management.FederationGroupMembership) error
```
wiliamsouza, line 25: `UpdateMembership(context.Context, *emtity.Membership) error`?

pkg/common/entity/membership.go
thisab: You are right

pkg/server/datastore/datastore.go
thisab: by doing that you would need to add ctx too:
CreateMember(ctx context.Context, member *entity.Member) (*entity.Member, error)

pkg/server/datastore/datastore.go
thisab: Doing that requires adding ctx. If so, then we would need to do that for all or none...
RetrieveMemberByID(ctx context.Context, member entity.Member) (*entity.Member, error)?

pkg/server/datastore/datastore.go
wiliamsouza: I like the idea to do definitions more explicit as possible so I like to name the parameters.

pkg/server/datastore/datastore.go
wiliamsouza: I like the idea to use named parameters for all definition.

pkg/server/datastore/sqlstore/model.go
```diff
@@ -0,0 +1,50 @@
...
+	TrustDomain   string
+	Memberships   []Membership   `gorm:"constraint:OnDelete:CASCADE;"`
+	Relationships []Relationship `gorm:"constraint:OnDelete:CASCADE;"`
+	TrustBundles  []TrustBundle  `gorm:"constraint:OnDelete:CASCADE;"`
+}
+
+type Membership struct {
+	Model
+	JoinToken string `gorm:"uniqueIndex"`
+	MemberID  uint
```
wiliamsouza, line 27: The type here would not be of the type `uuid.UUID`

pkg/server/datastore/sqlstore/model.go
```diff
@@ -0,0 +1,50 @@
...
+	TTL       uint
+}
+
+type Relationship struct {
+	Model
+	MemberID         uint // Implicit Foreign Key and also the SourceID for the relationship
```
wiliamsouza, line 33: The type here would not be of the type `uuid.UUID`

pkg/server/datastore/sqlstore/model.go
```diff
@@ -0,0 +1,50 @@
...
+	TargetMemberID   uint
```
wiliamsouza, line 34: The type here would not be of the type `uuid.UUID`

pkg/server/datastore/sqlstore/model.go
```diff
@@ -0,0 +1,50 @@
...
+	Status           string
+	RelationshipType string
+	TTL              uint
+}
+
+type TrustBundle struct {
+	Model
+	MemberID    uint //Implicit Foreign Key
```
wiliamsouza, line 42: The type here would not be of the type `uuid.UUID`

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,34 @@
...
+	entity "github.com/HewlettPackard/galadriel/pkg/common/entity"
+	"github.com/google/uuid"
+	"github.com/spiffe/spire/proto/spire/common"
+)
+
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(ctx context.Context, member *entity.Member) (*entity.Member, error)
+	CreateMembership(ctx context.Context, membership *entity.Membership) (*entity.Membership, error)
+	CreateTrustBundle(ctx context.Context, trustBundle *common.Bundle, memberID uint) (*common.Bundle, error)
```
wiliamsouza, line 15: The type should change to `memberID uuid.UUID`?

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,34 @@
...
+	CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceMemberId uint, targetMemberId uint) (*entity.Relationship, error)
```
wiliamsouza, line 16: The parameters `sourceMemberID `and  `targetMemberID ` have a different type of `uuid.UUID` having the following 
 signature: 
 `CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceMemberID, targetMemberID uuid.UUID) (*entity.Relationship, error)`

pkg/common/entity/relationship.go
```diff
@@ -0,0 +1,26 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/spiffe/spire/proto/spire/common"
+)
+
+type BundleEndpointType string
+
+type Relationship struct {
```
wiliamsouza, line 14: Missing the fields `sourceMemberID `and `targetMemberID` type `uuid.UUID` on this struct they exist in `CreateRelationship` method.

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,34 @@
...
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(ctx context.Context, member *entity.Member) (*entity.Member, error)
+	CreateMembership(ctx context.Context, membership *entity.Membership) (*entity.Membership, error)
+	CreateTrustBundle(ctx context.Context, trustBundle *common.Bundle, memberID uuid.UUID) (*common.Bundle, error)
+	CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceMemberID uuid.UUID, targetMemberID uuid.UUID) (*entity.Relationship, error)
+	RetrieveMembershipsByID(ctx context.Context, memberID uuid.UUID) (*[]entity.Membership, error)
+	RetrieveRelationshipsByID(ctx context.Context, memberID uuid.UUID) (*[]entity.Relationship, error)
+	RetrieveTrustBundlesByID(ctx context.Context, memberID uuid.UUID) (*[]common.Bundle, error)
+	RetrieveMemberByID(ctx context.Context, member entity.Member) (*entity.Member, error)
```
wiliamsouza, line 20: `RetrieveMemberByID(ctx context.Context, id uuid.UUID) (*entity.Member, error)`

pkg/common/entity/relationship.go
thisab: Atually missing the TargetMemberID. The source is the RelationshipID. Otherwise it would be redundant.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/HewlettPackard/galadriel/pkg/common/entity"
+	"gorm.io/gorm"
+)
+
+const (
+	// SQLite database type
+	SQLite = "sqlite3"
+)
+
+type Plugin struct {
```
mchurichi, line 16: I'd suggest using a more explicit name here, something like `SQLStore`

mchurichi: also, we could use [this](https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance) to verify DataStore interface compliance

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,34 @@
...
+	RetrieveMembershipByID(ctx context.Context, membershipID uuid.UUID) (*entity.Membership, error)
+	RetrieveRelationshipBySourceandTargetID(ctx context.Context, sourceMemberID uuid.UUID, targetMemberID uuid.UUID) (*entity.Relationship, error)
+	RetrieveTrustbundleByMemberID(ctx context.Context, memberID uuid.UUID) (*common.Bundle, error)
+	UpdateMember(ctx context.Context, member entity.Member) error
```
wiliamsouza, line 24: It can be interesting to return the updated entity as the return value form all the update methods cause this make test this more easy:

`UpdateMember(ctx context.Context, member *entity.Member) (*entity.Member, error)`

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,34 @@
...
+	UpdateMembership(ctx context.Context, membership entity.Membership) error
```
wiliamsouza, line 25: It's a good to pass parameters as reference we can save some memory doing that:
`UpdateMembership(ctx context.Context, membership *entity.Membership) (entity.Membership, error)`

pkg/server/datastore/datastore.go
thisab: But should the test trust the return value of the function?

pkg/common/entity/federationgroups.go
```diff
@@ -0,0 +1,15 @@
+package entity
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type FederationGroup struct {
+	FederationGroupID uuid.UUID
```
wiliamsouza, line 10: Rename `FederationGroupID ` to `ID` we end up with a more short use `FederationGroup.ID ` instead of `FederationGroup.FederationGroup.ID`?

pkg/common/entity/member.go
```diff
@@ -0,0 +1,16 @@
+package entity
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type Member struct {
+	MemberID    uuid.UUID
```
wiliamsouza, line 10: MemberID -> ID?

pkg/common/entity/federationgroups.go
thisab: I might agree, but I believe I was asked to chenge this before.. @mchurichi what do you think?

pkg/server/datastore/sqlstore/sqlstore.go
thisab: Hmm, I think Spire uses something similar to Plugin... @wiliamsouza

pkg/common/entity/federationgroups.go
mchurichi: mmm not sure who asked that, but I agree with @wiliamsouza

thisab: ok

pkg/common/entity/membership.go
```diff
@@ -0,0 +1,17 @@
+package entity
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type Membership struct {
+	MembershipID      uuid.UUID //SourceMemberID
```
wiliamsouza, line 10: MembershipID -> ID?
The idea is to use the Membership ID as source member ID? This implies in synchronize the two IDs? I think this can lead us to a limitation of data model as ID must be unique.

pkg/server/datastore/datastore.go
mchurichi: the test should not trust the return value, it must instead validate that the return value is the same as the member passed in (plus any additional added data in the member body, like the ID, CreatedDate, etc)

thisab: Therefore, the test should check the value stored. Now, if returning an entity in an update function is useful, then it should return it and the entity should also be checked.

wiliamsouza: What I meant is that returning the entity might be useful for the API, the Updated_At could be used, for example.

thisab: Oh, ok, Agreed then @wiliamsouza

pkg/server/datastore/sqlstore/sqlstore.go
mchurichi: Mmm yes, it does. I wonder if this is related to the plugin library they use. Anyways I'd still have an explicit naming here.

wiliamsouza: I agree with a more explicit name and interface compliance is a plus:

`var _ datastore.Datastore = &sqlstore.SqlStore{}`

pkg/common/entity/relationship.go
```diff
@@ -0,0 +1,26 @@
...
+	ID                    uuid.UUID
+	TrustDomain           spiffeid.TrustDomain
+	BundleEndpointURL     *url.URL
+	BundleEndpointProfile BundleEndpointType
+	TrustDomainBundle     *common.Bundle
+
+	// Fields only used for 'https_spiffe' bundle endpoint profile
```
jufantozzi, line 21: I think this comment is miss-leading, as we'll [always need](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Federation.md#522-spiffe-authentication-https_spiffe) to point to some bundle endpoint, regardless of the profile.

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,31 @@
...
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(ctx context.Context, member *entity.Member) (*entity.Member, error)
+	CreateMembership(ctx context.Context, membership *entity.Membership) (*entity.Membership, error)
+	CreateTrustBundle(ctx context.Context, trustBundle *common.Bundle, memberID uuid.UUID) (*common.Bundle, error)
+	CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceMemberID uuid.UUID, targetMemberID uuid.UUID) (*entity.Relationship, error)
+	RetrieveMemberByID(ctx context.Context, memberID uuid.UUID) (*entity.Member, error)
+	RetrieveMembershipByID(ctx context.Context, membershipID uuid.UUID) (*entity.Membership, error)
+	RetrieveRelationshipBySourceandTargetID(ctx context.Context, sourceMemberID uuid.UUID, targetMemberID uuid.UUID) (*entity.Relationship, error)
+	RetrieveTrustBundleByID(ctx context.Context, trustID uuid.UUID) (*common.Bundle, error)
```
wiliamsouza, line 20: `RetrieveTrustBundleByID(ctx context.Context, id uuid.UUID) (*common.Bundle, error)`?

jufantozzi: my take: `bundleID`

but also wondering what would be or if we would have any use case that justify this method...
this goes for all bundle operations

maxlambrecht:
> This is super picky but wanted to bring up a thing we were talking about with @jufantozzi just yesterday. It's about `ID` vs `Id`. There are a few forms for this:
> 
>     * `Id`/`ID` as in a table primary key
> 
>     * `MemberId`/`MemberID` as a reference to another table or entity's id
> 
>     * `RetrieveMemberById`/`RetrieveMemberByID` as a public API of the DataStore
> 
> 
> I don't have a strong opinion but I do believe we must choose a convention and stick to it.
> 
> Thoughts?

I'm in favor of following these rules: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

So, always `ID`

pkg/server/datastore/datastore.go
thisab: So, as we discussed. There is an open thread regarding the TrustBundle table.
If the relation is 1:1, I think this wouldnt be needed @jufantozzi

pkg/server/datastore/sqlstore/model.go
thisab: So @jufantozzi should we assume 1:1 relation from Member to TrustBundle?

pkg/common/entity/federationgroups.go
```diff
@@ -0,0 +1,8 @@
+package entity
```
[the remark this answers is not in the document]
thisab: I dont think so. What do you think @wiliamsouza ?

pkg/common/entity/relationship.go
thisab: This definition came from Spire. Thinking whether we would need those three  variables: BundleEndpointURL BundleEndpointProfile TrustDomainBundle. We already have a TrustBundle entity

thisab: @jufantozzi if not needed, we could just remove those variables

mchurichi: Both `BundleEndpointURL` and `BundleEndpointProfile` are needed for both `https_spiffe` and `https_web`.
`TrustDomainBundle` is needed for `https_spiffe` (this may be *a different bundle*  than the one we store in the TrustDomainBundle table, so we'd want to keep it here).
`EndpointSPIFFEID` is needed for `https_spiffe`.

mchurichi: See how it's being implemented in the SPIRE broker layer https://github.com/HewlettPackard/galadriel/pull/42/files#diff-7ee238c14b3184487c447a36bb1187b6417ed15e917e78bcde3703cbe14bfe6cR9

mchurichi: actually, the comment is accurate, `EndpointSPIFFEID` is only used on https_spiffe, so we may want to consider putting it in a HTTPSSpiffeBundleEndpointProfile as in the link above

pkg/server/datastore/sqlstore/model.go
mchurichi: My take on this, at least for the PoC: let's stick to a 1:1 Member -> TrustBundle entity and create a new TrustBundle record and delete the old one on rotation. We can later decide to whether keep the history or merge the TrustBundle table back to the Members table.

maxlambrecht: @mchurichi If we store the trust bundle as it comes in the response from the SPIRE Server we will have the old bundles CAs, thus storing previously fetched bundles would be redundant and add complexity. 

Regarding CA TTL, my take is that we should follow the [SPIFFE Federation spec](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Federation.md#41-adding-and-removing-keys) and poll for updates using the `spiffe_refresh_hint`.

Also from the SPIFFE spec:

```
Clients of the bundle endpoint SHOULD store the latest SPIFFE bundle each time it is retrieved. The sequence number field of the trust bundle SHOULD be used when comparing the freshness or ordering of two trust bundles.
```

pkg/common/entity/federationgroups.go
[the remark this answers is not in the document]
maxlambrecht: +1 for keeping them in `entity`, `common` is not as good as a package name as it doesn't convey any meaning about it contents

pkg/server/datastore/sqlstore/model.go
mchurichi: Does SPIRE keep old bundles in the SPIRE Server?
Anyways I think here we are discussing not about the previous bundle but older ones, like, let's say, 10 or more rotations before, that are no longer in the bundle coming from the SPIRE Server AFAIK

mchurichi: And whenever we say "TrustBundle" we talk about storing it as it comes from the Server, no stripping or any manipulation at all, let's SPIRE handle that.

pkg/common/entity/membership.go
```diff
@@ -0,0 +1,17 @@
+package entity
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type Membership struct {
+	ID                uuid.UUID //SourceMemberID
```
maxlambrecht, line 10: `ID` is the `SourceMemberID`?

thisab: @maxlambrecht what do you suggest? perhaps moving the entity to another folder?

mchurichi: I believe this is kinda confusing, what's the PK? this ID? I'd rather have an independent ID for the table and a SourceMemberID and TargetMemberID

pkg/common/entity/relationship.go
jufantozzi: My bad! I missed the point of this variable, was thinking about EndpointURL.

pkg/common/entity/federationgroups.go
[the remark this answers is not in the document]
jufantozzi: Max pointed out a good reason to keep it like this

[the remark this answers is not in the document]
thisab: ok then

pkg/common/entity/membership.go
thisab: Ok. We could have an independent ID for that in entity. Makes sense.

thisab: @maxlambrecht, @mchurichi actually SourceMemberID should be a property of Relationship, not Membership. So, Ive fixed that and made the variable SourceMemberID more explicit.

mchurichi: Right, good point

pkg/server/datastore/sqlstore/model.go
maxlambrecht: SPIRE keeps "old" bundles for certain time until they are stripped after a certain number of rotations.

pkg/common/entity/membership.go
thisab: Fixed.

pkg/server/datastore/sqlstore/model.go
```diff
@@ -0,0 +1,62 @@
...
+	Memberships   []Membership   `gorm:"constraint:OnDelete:CASCADE"`
+	Relationships []Relationship `gorm:"constraint:OnDelete:CASCADE;foreignKey:SourceMemberID"`
+	TrustBundles  TrustBundle    `gorm:"constraint:OnDelete:CASCADE"`
+}
+
+type Membership struct {
+	Model
+	JoinToken string    `gorm:"unique_index"`
+	MemberID  uuid.UUID `gorm:"type:uuid"`
+	TTL       uint
```
mchurichi, line 39: wasn't there a Status field for the Membership?

pkg/server/datastore/sqlstore/model.go
mchurichi: ok, so looks like we won't need to keep any old bundle at all since SPIRE is already taking care of that, right? If so, I'd say let's add a TrustBundle string field in the Members table, thoughts?

pkg/server/datastore/sqlstore/sqlite.go
```diff
@@ -0,0 +1,46 @@
...
+		u.Opaque, u.Path = u.Path, ""
+	case u.Scheme != "file":
+		return "", fmt.Errorf("unsupported scheme %q", u.Scheme)
+	}
+
+	q := u.Query()
+	q.Set("_foreign_keys", "ON")
+	q.Set("_journal_mode", "WAL")
+	u.RawQuery = q.Encode()
+	return u.String(), nil
```
mchurichi, line 45:
```suggestion
	u.RawQuery = q.Encode()
	
	return u.String(), nil
```

pkg/server/datastore/sqlstore/sqlite.go
```diff
@@ -0,0 +1,46 @@
...
+		return "", err
+	}
+
+	switch {
+	case u.Scheme == "":
+		u.Scheme = "file"
+		u.Opaque, u.Path = u.Path, ""
+	case u.Scheme != "file":
+		return "", fmt.Errorf("unsupported scheme %q", u.Scheme)
+	}
```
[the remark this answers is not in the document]
mchurichi: btw why are we switching and emptying u.Opaque and u.Path?

pkg/server/datastore/sqlstore/model.go
thisab: Not inside Membership. But there was one inside the Relationship struct.

pkg/server/datastore/sqlstore/model.go
thisab: @mchurichi if we follow that, it would take away the need for a specific table for the TrustBundle right? If so, I think it simplifies it further.

pkg/server/datastore/sqlstore/sqlite.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"net/url"
+
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+type sqliteDB struct{}
+
+func (s sqliteDB) connect(connectionString string) (db *gorm.DB, err error) {
```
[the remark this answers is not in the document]
thisab: those values are used inside the function itself. To change it I would need to declare them inside the function. If that is what is intended, ok.

pkg/server/datastore/sqlstore/sqlite.go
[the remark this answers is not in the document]
thisab: Oh, the reasoning is the same: [here](https://github.com/spiffe/spire/blob/main/pkg/server/datastore/sqlstore/sqlite.go). perhaps I should have kept the comment.

pkg/server/datastore/sqlstore/sqlite.go
thisab: just add an empty line?

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/google/uuid"
+	"github.com/spiffe/spire/proto/spire/common"
+)
+
+// DataStore defines the data storage interface.
+type DataStore interface {
+	CreateMember(ctx context.Context, member *entity.Member) (*entity.Member, error)
+	CreateMembership(ctx context.Context, membership *entity.Membership) (*entity.Membership, error)
+	CreateTrustBundle(ctx context.Context, trustBundle *common.Bundle, trustBundleID uuid.UUID) (*common.Bundle, error)
+	CreateRelationship(ctx context.Context, relationship *entity.Relationship, sourceMemberID uuid.UUID, targetMemberID uuid.UUID) (*entity.Relationship, error)
```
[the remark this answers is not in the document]
thisab: Well.. I thought a bit about it, as all of them have ID fields, we could resume all CRUD operations to use ByID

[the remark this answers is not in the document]
mchurichi: what you mean with "to use ByID"?

pkg/common/entity/relationship.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+type Relationship struct {
+	ID                    uuid.UUID
+	SourceMemberID        uuid.UUID
+	TargetMemberID        uuid.UUID
+	TrustDomain           spiffeid.TrustDomain
+	BundleEndpointURL     *url.URL
+	BundleEndpointProfile BundleEndpointProfile
+	TrustDomainBundle     *spiffebundle.Bundle
```
jufantozzi, line 23: Which Member does this bundle refer to? I think the trust bundle should be part of `Membership`

pkg/common/entity/relationship.go
```diff
@@ -0,0 +1,27 @@
...
+	EndpointSPIFFEID      spiffeid.ID
```
jufantozzi, line 24: Same consideration here, but maybe this one can be part of `Member` as it's not time sensitive

pkg/server/datastore/datastore.go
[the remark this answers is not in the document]
thisab: Ive uploaded the suggested change. What I meant is that instead of using RetrieveRelationshipBySourceMemberTargetMemberID we use RetrieveRelationshipByID

pkg/common/entity/relationship.go
mchurichi: This is the bundle used to bootstrap the federation with the target member, it could be either the other Member's bundle or a completely different one (see ), we need to keep it here.
See _not self-serving endpoint_ (2nd bullet) [here](https://github.com/spiffe/spiffe/blob/27b59b81ba8c56885ac5d4be73b35b9b3305fd7a/standards/SPIFFE_Federation.md#5222-endpoint-parameters)

mchurichi: we could rename this to `EndpointTrustBundle` to make it clearer

pkg/server/datastore/sqlstore/model.go
jufantozzi: I think we need it if we want to be able to revoke a Membership

pkg/server/datastore/sqlstore/model.go
jufantozzi: yes, from this discussion seems like we shouldn't need a table for Trust Bundles. 

I think it would be conceptually better suited to have it in the Membership, as these are time-sensitive entities (membership and bundle)

pkg/common/entity/relationship.go
mchurichi: Be aware that the SPIFFE ID in Member is the one that is configured in the SPIRE Server.
The EndpointSPIFFEID here relates to the SVID that the Target Member will present when establishing the connection. It could be either the SPIFFEID of the SPIRE Server (in which case it will match the one in the Members table) or a completely different one (see again _not self-serving endpoint_).
We should keep this here, and I think this name makes sense as it matches the FederationRelationship proto in the TrustBundle API (a.k.a. Federation API)

pkg/server/datastore/sqlstore/sqlite.go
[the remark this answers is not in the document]
mchurichi: Ah ok, it makes sense now. Yes, please keep the comment there as it's not easy to reason about at first sight. Also please use two ifs instead of a switch as it is easier to read IMO.

pkg/server/datastore/sqlstore/model.go
thisab: Ok, inserted on both then

pkg/common/entity/relationship.go
mchurichi: correct

mchurichi: we were talking with @jufantozzi about the possibility of prefixing all fields that refer to the Target with the `Target` prefix to make it easier to reason about... `TargetBundleEndpointURL`, `TargetEndpointTrustBundle`, etc.
Thoughts?

pkg/server/datastore/sqlstore/model.go
thisab: @jufantozzi Membership ties a Member to a FedGroup (N:M). By inserting the Bundle inside the Membership that could imply that for each distinct FederationGroup, a single member can have multiple TrustBundles (distinct or not). What Im conveying is that if a TrustBundle is unique to a Member, maybe it could be its attribute. And that might make it easier if we want to create a history table attached to Member later on.

pkg/common/entity/relationship.go
jufantozzi: As we discussed offline, I had a miss-conception of directionality of relationships for Galadriel. This makes sense now 😊

pkg/common/entity/relationship.go
thisab: 100% agreed, makes it even more explicit

thisab: TargetTrustDomain also?

thisab: Thinking about it: Is TargetTrustDomain the same as the TrustDomain from the Member referenced by TargetMemberID in Relationship entity?

If so, we may not need it here, just get the value by using TargetMemberID to search inside the Member table.

mchurichi: yes
```golang
TargetTrustDomain
TargetBundleEndpointURL
TargetBundleEndpointProfile
TargetTrustBundle // not TrustDomainBundle
TargetEndpointSPIFFEID
```

pkg/server/datastore/sqlstore/model.go
jufantozzi: Good point! Having it in member sounds good

pkg/common/entity/relationship.go
mchurichi: edited

pkg/server/datastore/sqlstore/model.go
thisab: Understood. So shall we remove the TrustBundle table and related crud functions?

thisab: @mchurichi changed the TrustBundle to be an attribute of Member and ajusted the code.

PR #41, opened 2022-08-29
[Harvester] Fetch the current Trust Bundle from a SPIRE Server

mchurichi:
Signed-off-by: Maximiliano Churichi <mchurichi@gmail.com>

**Pull Request check list**

- [x] Proper tests/regressions included?
- [ ] Documentation updated?

**Affected functionality**
N/A

**Description of change**
Implements the SPIRE plugin interface to get a SPIRE Server bundle via UDS.

Test steps:
- Run a SPIRE Server
- Change the default config file of the harvester and set `spire_socket_path` to the UDS of the SPIRE Server
- Run `go run cmd/harvester/main.go spire-bundle`

**Which issue this PR fixes**
N/A

PR #42, opened 2022-08-30
[Harvester] List federation relationships of a SPIRE Server

mchurichi:

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
N/A

**Description of change**
Implements the SPIRE plugin interface to list the configured federation relationships on a SPIRE Server via UDS.

Test steps:
- Run a SPIRE Server
- Change the default config file of the harvester and set `spire_socket_path` to the UDS of the SPIRE Server
- Create some federation relationships with the following command: `curl -s https://gist.githubusercontent.com/mchurichi/474659c894b57f7411a970e74468aefd/raw/89fa531f88fb06a3980a9e324508e84ee2d5c179/relationships.json | bin/spire-server federation create -socketPath <path to the UDS> -data -`
- Run `go run cmd/harvester/main.go spire-list-federations`

**Which issue this PR fixes**


cmd/harvester/cli/cli.go
```diff
@@ -2,11 +2,14 @@ package cli
```
mchurichi, line 2: The changes introduced into this file are only intended to facilitate the testing of this feature and will be removed before merging this PR

pkg/harvester/spire/spire.go
```diff
@@ -3,36 +3,82 @@ package spire
...
 }
 
-func (s *LocalSpireServer) GetFederationRelationships(ctx context.Context) ([]spireapi.FederationRelationship, error) {
-	return nil, errors.New("not implemented")
+type clientMaker func(*grpc.ClientConn) (client, error)
+
+func dialSocket(ctx context.Context, path string, makeClient clientMaker) (client, error) {
+	var target string
+
+	if filepath.IsAbs(path) {
```
maxlambrecht, line 56: This logic can be moved to an utility method and create a UnixAddr, looks more robust. 

```
func GetUnixAddrWithAbsolutePath(path string) (*net.UnixAddr, error) {
	pathAbs, err := filepath.Abs(path)
	if err != nil {
		return nil, fmt.Errorf("failed to get absolute path: %w", err)
	}

	return &net.UnixAddr{
		Name: pathAbs,
		Net:  "unix",
	}, nil
}
```

jufantozzi (commented):
🍨 🥄 
Just left some questions that came to mind.

pkg/harvester/spire/spire.go
mchurichi: sounds good! I'll implement this in #41  and rebase once merged

cmd/harvester/cli/cli.go
```diff
@@ -22,22 +25,46 @@ func NewHarvesterCLI() *HarvesterCLI {
...
+		c.logger.Error("Error getting federation relationships:", err)
+		return 1
+	}
+	if len(feds) == 0 {
+		c.logger.Info("No federation relationships found")
+		return 0
+	}
+
+	for _, fed := range feds {
+		c.logger.Info(fmt.Sprintf("Trust Domain: %s", fed.TrustDomain))
```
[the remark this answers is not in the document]
mchurichi: Right, I don't think this is how we are going to print output. We'll probably rely on some wrapper around fmt.Print, I think spire does something like that but I haven't looked deep into it.

pkg/harvester/spire/spire.go
```diff
@@ -3,36 +3,82 @@ package spire
...
+type client interface {
+	TrustDomainClient
+}
+
+var dialFn = dialSocket
+var grpcDialContext = grpc.DialContext
+
 func NewLocalSpireServer(socketPath string) SpireServer {
-	return &LocalSpireServer{
+	client, err := dialFn(context.Background(), socketPath, makeSpireClient)
```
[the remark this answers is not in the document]
mchurichi: good catch, we should be using the same ctx

pkg/harvester/spire/spire.go
mchurichi: I've been looking into how to feed the grpc Dial process with a UnixAddr instead of a string and I can say that it's pretty tricky. It heavily relies on a `target` string (a uds, dns, [etc](https://github.com/grpc/grpc/blob/master/doc/naming.md)), a [resolver.Resolver](https://pkg.go.dev/google.golang.org/grpc@v1.48.0/resolver#Resolver) and many layers below to help shape the final connection details (the address is passed as a string until well below).

I believe this isn't worth the effort. If the concern is the potential flakiness of relying on a string from the user we can do some sanity checks before passing it to the calling mechanism. Thoughts?

mchurichi: We could also move this logic to get the final UDS path to a helper method if that improves the readability.

jufantozzi:
Thanks @mchurichi, I'm ready to approve this, we just need to resolve the conflicts

mchurichi:
> Thanks @mchurichi, I'm ready to approve this, we just need to resolve the conflicts

sorry, my bad, I tried to re-request your request in #41, not in this one 😆

mchurichi:
I will resolve the conflicts while waiting to hear back from @maxlambrecht

cmd/harvester/cli/spire.go
```diff
@@ -0,0 +1,56 @@
+package cli
```
mchurichi, line 1: Again, this file is only intended to facilitate testing via go run cmd/harvester/main.go spire-bundle. It will be removed before merging this PR

jufantozzi: Btw, do you think `api.go` in an appropriate name for hosting the `run` command? IMO we should call it `run.go`

mchurichi: Agreed. The closest the files and components' names are to the command a user will run, the better (one of the premises of DDD).

jufantozzi (dismissed):

maxlambrecht (approved):

pkg/harvester/spire/spire.go
mchurichi: working on this, I'll push the changes shortly

mchurichi:
Closing for now as it's out of scope for the PoC

PR #43, opened 2022-08-30
[Draft] API Definition review

jufantozzi:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

This PR is a draft to promote discussion around updating our API definitions. 
The goal is to review the spec so it aligns with latest definition we have while also reflecting goals for our PoC.
 

**Which issue this PR fixes**


jufantozzi:
Hey @thisab, just realized that William Souza does not have access to this repo. Would you please share his user so I can add him to this discussion?

thisab:
> Hey @thisab, just realized that William Souza does not have access to this repo. Would you please share his user so I can add him to this discussion?

Sure, here it is: https://github.com/wiliamsouza

wiliamsouza:
I whould like to discuss the use of `HTTP` method `PUT` on API definition. Should we keep it or change to use `PATCH`?
`PATCH` is for making partial changes to an existing resource other hand `PUT` is used for replace an existing resource content[1]. 

[1] https://josipmisko.com/posts/patch-vs-put-rest-api

spec/api/management.yaml
```diff
@@ -14,29 +14,30 @@ components:
...
+#        - name
     FederationGroup:
-      # A Federation Group is a group of SpireServers (representation of the bridge)
+      # A Federation Group represents a bridge for its Members,
+      # allowing relationships to be established
       type: object
       properties:
         id:
           type: integer
           format: int64
```
thisab, line 37: @jufantozzi could we use format: uuid to store the ID?

thisab, line 37:
```suggestion
          type: string
          format: uuid
```
@jufantozzi maybe we could use format: uuid to store the ID?

thisab (commented):
Adding my 2 cents on a topic discussed earlier.

- There are some other API definitions (/organizations) that we may strip for the PoC?
- Will federationGroupMemberships become just Memberships ?

mchurichi:
> I whould like to discuss the use of `HTTP` method `PUT` on API definition. Should we keep it or change to use `PATCH`? `PATCH` is for making partial changes to an existing resource other hand `PUT` is used for replace an existing resource content[1].
> 
> [1] https://josipmisko.com/posts/patch-vs-put-rest-api

PATCH is what people want 99% of the time, I'd suggest to use it instead of PUT

jufantozzi:
> I wasn't aware we are stripping Organization. Is it for PoC purposes?

AFAIK by stripping I think he meant we're removing.

mchurichi:
> > I wasn't aware we are stripping Organization. Is it for PoC purposes?
> 
> AFAIK by stripping I think he meant we're removing. Was your question regarding it's removal?

Got it, and I wasn't aware of that decision.

More food for thought:
- `/federationRelationships` -> `/relationships`
- `/spireServers`  -> `/members`

thisab:
> > I wasn't aware we are stripping Organization. Is it for PoC purposes?
> 
> AFAIK by stripping I think he meant we're removing. Was your question regarding it's removal?

Edited my previous comment, but yes and for PoC purposes only.
Maybe: `/federationGroupMemberships` -> `/membership`

spec/api/management.yaml
mchurichi: how does uuid's translates to the different dbms?

thisab: I can say that [gorm](https://medium.com/@the.hasham.ali/how-to-use-uuid-key-type-with-gorm-cc00d4ec7100) supports it. [MongoDB](https://www.mongodb.com/docs/manual/reference/method/UUID/) seems too

jufantozzi:
Made the updates already agreed on here. There one concern that came to mind when reviewing, and this is something I also asked on the design doc: How will the trust bundle endpoint work? will we actually allow for those to be manually manipulated?

mchurichi:
> Made the updates already agreed on here. There's one concern that came to mind when reviewing, and this is something I also asked on the design doc: How will the trust bundle endpoint work? will we actually allow for those to be manually manipulated?

What do you mean by manually manipulated? To allow an admin user to change/delete Trust Bundles? that doesn't feel quite right.

AFAIK the spire server does not provide a method to change/update its own current bundle. You can only append a new one, but not delete or change one that is already there.

jufantozzi:
> What do you mean by manually manipulated? To allow an admin user to change/delete Trust Bundles? that doesn't feel quite right.

I'm referring to this [very hidden path](https://github.com/HewlettPackard/galadriel/pull/43/files#diff-df80259918c1ba12005e4825cf76acd82233f079d30c7e88d84489bd7acc81fcR727), I'm not sure either what was the idea behind it but seems off to me.

> AFAIK the spire server does not provide a method to change/update its own current bundle. You can only append a new one, but not delete or change one that is already there.

that's my understanding as well.

mchurichi:
> > What do you mean by manually manipulated? To allow an admin user to change/delete Trust Bundles? that doesn't feel quite right.
> 
> I'm referring to this [very hidden path](https://github.com/HewlettPackard/galadriel/pull/43/files#diff-df80259918c1ba12005e4825cf76acd82233f079d30c7e88d84489bd7acc81fcR727), I'm not sure either what was the idea behind it but seems off to me.

IMO that shouldn't be allowed.

jufantozzi:
Closing this for now, we should be re-visiting in the future

PR #47, opened 2022-09-09
Refactoring codebase

maxlambrecht:

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

Refactor to organize the responsibilities of the entities across the codebase: 

- Created a Server and Harvester entities that start Endpoints with TCP and UDS addresses. 
- Created and moved around new Config structs for handling the configurations where they need to be (e.g., CLI configs reading from config files are now in the CMD folder)
- Refactored Catalog to have one for the Harvester and other for the Server (although theses objects will not be needed for the PoC implementation)
- Converted to structs some interfaces that are not good abstractions (e.g., the HarvesterController).
- Added a MemStore to be used as long as we don't have an actual datastore implementation, and probably enough for a PoC. 
- Added the `generate token` functionality, although there is some room for improvement.
- Added `logrus` for Logging instead of using a handmade Logger.
- Some other minor changes like adding or changing config options in the conf files. 

** This refactor may be breaking some functionality, although Server and Harvester are running **

**Which issue this PR fixes**


pkg/common/telemetry/names.go
```diff
@@ -11,6 +11,7 @@ const (
 	TrustBundle = "trust_bundle"
 	PackageName = "package_name"
 	Federation  = "federation"
+	Su
```
mgbcaio, line 14: why is this here ?

pkg/common/telemetry/names.go
```diff
@@ -26,12 +27,17 @@ const (
...
+
 	MetricsServer       = "metrics_server"
 	HarvesterController = "harvester_controller"
 
 	GaladrielServer = "galadriel_server"
 	HTTPApi         = "http_api"
 
-	ID = "id"
-	// spiffeID = "spiffeID"
+	// SubsystemName declares field for some subsystem name (an API, module...)
```
mgbcaio, line 39:
```suggestion
	// SubsystemName declares a field for some subsystem name (an API, module...)
```

cmd/harvester/cli/cli.go
```diff
@@ -1,22 +1,22 @@
...
-	"github.com/HewlettPackard/galadriel/pkg/common"
 	"github.com/HewlettPackard/galadriel/pkg/common/telemetry"
 )
 
 var (
 	RootCmd      = NewRootCmd()
 	cmdExecute   = RootCmd.Execute
 	HarvesterCmd = &HarvesterCLI{
-		logger: common.NewLogger(telemetry.Harvester),
+		log: logrus.WithField(telemetry.SubsystemName, telemetry.Harvester),
```
mgbcaio, line 14: oh nice!!! 💯

cmd/harvester/cli/config.go
```diff
@@ -0,0 +1,117 @@
...
+	ListenPort      int    `hcl:"listen_port"`
+	SocketPath      string `hcl:"socket_path"`
+	SpireSocketPath string `hcl:"spire_socket_path"`
+	ServerAddress   string `hcl:"server-address"`
+	LogLevel        string `hcl:"log_level"`
+}
+
+// ParseConfig reads a configuration from the Reader and parses it
+// to a cli.Config object setting the defaults for the missing values.
+func ParseConfig(config io.Reader) (*Config, error) {
```
mgbcaio, line 35: this function is the same present in the server, do you think we could extract it to only have it implemented once? (maybe put it inside a util package or something like that).

pkg/harvester/client/server.go
```diff
@@ -0,0 +1,39 @@
+package client
+
+import (
+	"context"
+	"errors"
+	"github.com/HewlettPackard/galadriel/pkg/common/telemetry"
+	"github.com/sirupsen/logrus"
+)
+
+// GaladrielServerClient represents a client to connecto to Galadriel Server
```
mgbcaio, line 10:
```suggestion
// GaladrielServerClient represents a client to connect to Galadriel Server
```

pkg/common/util/task.go
```diff
@@ -0,0 +1,53 @@
+package util
+
+import (
+	"context"
+	"fmt"
+	"runtime/debug"
+	"sync"
+)
+
+// RunTasks runs all the given tasks concurrently and waits for all of them to completed.
```
mgbcaio, line 10:
```suggestion
// RunTasks runs all the given tasks concurrently and waits for all of them to be completed.
```

mgbcaio (changes requested):
This is fantastic, Max L. thanks for the refactoring. I left some nit comments and other than that we have some GitHub actions lint failures if you could check those =)

cmd/server/cli/config.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	return newConfig(configBytes)
+}
+
+// NewServerConfig creates a server.Config object from a cli.Config.
+func NewServerConfig(c *Config) (*server.Config, error) {
+	sc := &server.Config{}
+
+	ip := net.ParseIP(c.Server.ListenAddress)
```
mchurichi, line 51: This should allow either a hostname or IP address to provide flexible interface binding

cmd/server/cli/config.go
```diff
@@ -0,0 +1,101 @@
...
+
+func newConfig(configBytes []byte) (*Config, error) {
+	var config Config
+
+	if err := hcl.Decode(&config, string(configBytes)); err != nil {
+		return nil, fmt.Errorf("unable to decode configuration: %v", err)
+	}
+
+	if config.Server == nil {
+		return nil, errors.Wrap(errors.New("configuration file is empty"), "bad configuration")
```
mchurichi, line 77: AFAIK the idea behind wrapping an error is to allow callers to assert if a wrapped error is a specific error implementation, but since we are not exposing "bad configuration" here, not sure if this makes sense

cmd/server/cli/run.go
```diff
@@ -21,11 +25,33 @@ func NewRunCmd() *cobra.Command {
...
+
+			c, err := ParseConfig(configFile)
+			if err != nil {
+				return err
+			}
+
+			sc, err := NewServerConfig(c)
+			if err != nil {
+				return err
+			}
```
mchurichi, line 42: does it make sense to have file reading and config mapping here? it feels we'll be reusing this

pkg/server/endpoints/endpoints.go
```diff
@@ -0,0 +1,111 @@
...
+		err = e.DataStore.CreateJoinToken(ctx, t)
+		if err != nil {
+			e.Log.Errorf("failed to generate token: %v", err)
+			w.WriteHeader(500)
+			return
+		}
+
+		_, err = io.WriteString(w, t.Token)
+		if err != nil {
+			return
```
mchurichi, line 108: don't we need to return some kind of error if this fails?

pkg/server/datastore/datastore.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+)
+
+type DataStore interface {
+	CreateJoinToken(context.Context, *JoinToken) error
```
mchurichi, line 11: It's usually a good idea to have all create and update methods returning the newly created/updated object

pkg/harvester/endpoints/endpoints.go
```diff
@@ -0,0 +1,69 @@
...
+
+	select {
+	case err = <-errLocalServer:
+	case <-ctx.Done():
+		if err != nil {
+			fmt.Printf("error serving HTTP on uds: %v", err)
+		}
+		e.Log.Println("Stopping HTTP Server")
+		localServer.Close()
+		tcpServer.Close()
```
mchurichi, line 65: shouldn't we have these defer'ed after starting the servers? as a panic before reaching this point will prevent an appropriate cleanup

pkg/harvester/endpoints/endpoints.go
```diff
@@ -0,0 +1,69 @@
...
+		errLocalServer <- localServer.Serve(l)
+	}()
+
+	errTcpServer := make(chan error)
+	go func() {
+		errTcpServer <- tcpServer.Start(e.TCPAddress.String())
+	}()
+
+	select {
+	case err = <-errLocalServer:
```
mchurichi, line 58: remove, use, or
```suggestion
	case err = <-errLocalServer: //nolint:ineffassign
```

pkg/common/telemetry/names.go
maxlambrecht: Oops

cmd/harvester/cli/config.go
maxlambrecht: The functions are not the same, the return types are different, one is `harvester.Config` and the other is `server.Config`

pkg/harvester/catalog/catalog.go
```diff
@@ -1,11 +1,33 @@
...
-	"github.com/HewlettPackard/galadriel/pkg/harvester/spire"
+	"context"
+	"github.com/HewlettPackard/galadriel/pkg/common/telemetry"
+	"github.com/sirupsen/logrus"
 )
 
-type Catalog struct {
-	Spire  spire.SpireServer
-	Server server.GaladrielServer
+type Catalog interface {
```
jufantozzi, line 9: Can we please have a description for this interface? personally this was one of the things I found hard to understand the meaning of in SPIRE (still not comfortable as I'd like 😅 )

jufantozzi (commented):
Regarding POC scope, LGTM!

jufantozzi:
BTW regarding the datastore implementation, I can incorporate the token logic into the other PR to avoid conflicts

cmd/server/cli/config.go
maxlambrecht: That would be complicated, as the TCPAddr object uses an IP, if we accept a hostname we would need to look up the name to obtain the IP.

maxlambrecht:
I addressed some of the comments, still working through all the comments. And also I'm removing some components that won't be necessary.

cmd/server/cli/config.go
mchurichi: ok it makes sense, no reason to add this complexity for a PoC

jufantozzi:
more thoughts on the database: we can follow with the `in-mem` until we're done with the poc. Then we eventually start implementing the interface for the `sqlstore`, WDYT?

maxlambrecht:
> more thoughts on the database: we can follow with the `in-mem` until we're done with the poc. Then we eventually start implementing the interface for the `sqlstore`, WDYT?

yes, we can make the DB CRUD methods use some structure in memory to store the objects.

pkg/harvester/client/server.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (s *client) GetUpdates(ctx context.Context) ([]string, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (s *client) PushUpdates(ctx context.Context, updates []string) error {
+	return errors.New("not implemented")
+}
+
+func (s *client) GetMemberships(ctx context.Context) ([]string, error) {
```
jufantozzi, line 37: are we keeping memberships for POC?

cmd/server/cli/config.go
maxlambrecht: You are right, it doesn't make sense. 
It was not on purpose, I moved that method from another file and it already had that return statement.

pkg/harvester/client/server.go
maxlambrecht: removed

jufantozzi (dismissed):

mgbcaio (dismissed):

pkg/harvester/spire/spire.go
```diff
@@ -61,12 +63,12 @@ func dialSocket(ctx context.Context, path string, makeClient clientMaker) (clien
 	}
 	clientConn, err := grpcDialContext(ctx, target, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
-		return nil, fmt.Errorf("failed to dial API socket: %v", err)
+		return nil, fmt.Errorf("failed to dial API socket: %w", err)
```
mchurichi, line 66: not a huge concern, but wouldn't we leaking implementation details (internal components error) if we wrap these errors?

mchurichi: I think we should only wrap errors if it's an error we implement and expose

cmd/server/util/client.go
```diff
@@ -0,0 +1,46 @@
...
+	return serverClient{client: c}
+
+}
+
+type serverClient struct {
+	client *http.Client
+}
+
+func (c serverClient) GenerateJoinToken() (string, error) {
+	r, err := c.client.Get("http://unix/token")
```
mchurichi, line 35: where does this URL come from?

maxlambrecht: the Get method requires a well-formed URL with the Scheme and Host components, although that's irrelevant for a call on a Unix domain socket. Only the path "/token" is important

mchurichi (approved):

mgbcaio (approved):

PR #49, opened 2022-09-20
Adds SetFederatedBundle to a SPIRE Server

mchurichi:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**
Adds a method to set Federated Bundles in a SPIRE Server.

**Which issue this PR fixes**

### DIFF
--- a/pkg/harvester/spire/proto.go
+++ b/pkg/harvester/spire/proto.go
@@ -3,17 +3,20 @@ package spire
 import (
 	"crypto"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/spiffe/go-spiffe/v2/bundle/spiffebundle"
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	bundlev1 "github.com/spiffe/spire-api-sdk/proto/spire/api/server/bundle/v1"
 	apitypes "github.com/spiffe/spire-api-sdk/proto/spire/api/types"
+	"google.golang.org/grpc/codes"
 )
 
 func protoToBundle(in *apitypes.Bundle) (*spiffebundle.Bundle, error) {
 	if in == nil {
-		return nil, fmt.Errorf("bundle is empty")
+		return nil, errors.New("bundle is empty")
 	}
 
 	td, err := spiffeid.TrustDomainFromString(in.TrustDomain)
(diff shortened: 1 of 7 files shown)


jufantozzi (commented):
Didn't run any tests on it but so far looks good :)

mgbcaio (commented):
LGTM! 🌽

pkg/harvester/spire/spire.go
```diff
@@ -51,17 +52,24 @@ func (s *localSpireServer) GetBundle(ctx context.Context) (*spiffebundle.Bundle,
...
 type clientMaker func(*grpc.ClientConn) (client, error)
 
 func dialSocket(ctx context.Context, path string, makeClient clientMaker) (client, error) {
-	var target string
-
-	if filepath.IsAbs(path) {
-		target = "unix://" + path
-	} else {
-		target = "unix:" + path
+	uds, err := util.NormalizeUDSPath(path)
```
maxlambrecht, line 67: This path has already been normalized: https://github.com/HewlettPackard/galadriel/blob/75964dc06f8ea043d2f35eae2a42499f444a0775/cmd/harvester/cli/config.go#L47

I think instead of passing around an address as a string it should be a `net.Addr` as it is in the `harvester.Config` struct.

pkg/harvester/spire/proto.go
```diff
@@ -23,12 +26,12 @@ func protoToBundle(in *apitypes.Bundle) (*spiffebundle.Bundle, error) {
 
 	x509authorities, err := protoToX509Certificates(in.X509Authorities)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse x509 authorities: %v", err)
+		return nil, fmt.Errorf("failed to parse X.506 authorities: %v", err)
```
maxlambrecht, line 29: nit: `X.509`

mchurichi: I owe you 3 😅

pkg/harvester/spire/spire.go
mchurichi: Cool, I'll remove the normalization step.
Regarding the address, it needs to be a string as it's what the grpc library takes as an input. We can pass a net.Addr to this method, but it will need to be converted to a string with the scheme prefix `unix://`.

maxlambrecht: If the method receives a `net.Addr`, you can just pass `fmt.Sprintf("%s://%s", addr.Network(), addr.String())` to the grpc library.

mchurichi: sounds good, thanks!

maxlambrecht (approved):
🌭

jufantozzi (dismissed):
🚴‍♂️

mchurichi:
Thanks all for the reviews 🙌 
@maxlambrecht @jufantozzi could you please give it another ✅? I had to force push to fix the missing commit signatures and lost your approvals

jufantozzi (approved):

PR #52, opened 2022-09-29
 Implements harvester bundle endpoints

jufantozzi:
**Description of change**

Complete changelog:
* General cleanups and refactors:
    * CLI package (cleanup),
    * Common package and
    * Endpoints package (split in admin and harvester package).
    * Unused left-overs.
* Implements 2 new endpoints, `/bundle/sync` and  `/bundle`. 
* Creates basic testing make targets for the implemented endpoints. This should be improved and be leveraged by our pipeline at some point.


Makefile
```diff
@@ -140,6 +140,25 @@ test-unit:
...
+## Temporary targets to test sync APIs
+create-relationship:
+	./bin/galadriel-server create member -t td1
+	./bin/galadriel-server create member -t td2
+	./bin/galadriel-server create relationship -a td1 -b td2
+
+.PHONY: create-relationship
+
+run-harvester: create-relationship
+	token=`./bin/galadriel-server generate token -t td2 | ggrep -Po "(?<=Access Token:\s).*"`; \
```
jufantozzi, line 153:
```suggestion
	token=`./bin/galadriel-server generate token -t td2 | grep -Po "(?<=Access Token:\s).*"`; \
```

mgbcaio:
Excellent job, Juliano. I think this PR will have a lot of conflicts with the others, or the other PRs will have conflicts with this xD
I suggest we look into this to decide which implementation we are going to choose (since you removed the `draft` label)

jufantozzi:
> I suggest we look into this to decide which implementation we are going to choose
What implementation do you mean? the code for this PR should be complete by now

mgbcaio:
> > I suggest we look into this to decide which implementation we are going to choose
> 
> What implementation do you mean? the code for this PR should be complete by now

There are things you changed that were also changed in other PRs, that's pretty much it. for example the `addHandlers` method, just to avoid conflicts, either you or Maxi would need to change the file =)

Same for the paths in the `cmd/server/util/client.go` file and so on, map those conflicts would take a lot of time xD

mgbcaio:
> > I think this PR will have a lot of conflicts with the others, or the other PRs will have conflicts with this
> 
> My impression is that those conflicts will be minor and hopefully easy to handle. The changes introduced by #57 should take a more careful merge than #56 or #54, but nothing really complicated as same functions were not touched.

Fixing it now or later does not matter, as long as we do not have more work than we intend to. Just saying to keep an eye on this, since conflicts will occur.

jufantozzi:
> Fixing it now or later does not matter, as long as we do not have more work than we intend to. Just saying to keep an eye on this, since conflicts will occur.

Sometimes you can't avoid conflicts and thats okay too =) I did a small refactor because we had 2 http servers and handlers on a single package and it was getting clunky. Code was simply split into different packages, that shouldn't really be a big problem for maxi to work around. If I'm seeing this wrong, @mchurichi please feel free to drop in whenever.

mchurichi:
> 

Agree @jufantozzi, I didn't really add much, just a couple of new methods that can be easily carried over to the right file.

mgbcaio:
> > Fixing it now or later does not matter, as long as we do not have more work than we intend to. Just saying to keep an eye on this, since conflicts will occur.
> 
> Sometimes you can't avoid conflicts and thats okay too =) I did a small refactor because we had 2 http servers and handlers on a single package and it was getting clunky. Code was simply split into different packages, that shouldn't really be a big problem for maxi to work around. If I'm seeing this wrong, @mchurichi please feel free to drop in whenever.

yeah, of course, just bringing this up so that we can solve it together =P

jufantozzi:
Just solidifying the decided order of merges discussed offline with @mchurichi:

1. First we merge #54 and #57, which are easily rebasable into this one
2. Then we rebase and merge this PR
3. Then #56 can we be rebased with all the changes

pkg/server/api/harvester/handlers.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (s *echoServer) calculateBundleSync(
+	received common.FederationState,
+	current []*datastore.Relationship,
+	receivedID uuid.UUID) common.FederationState {
+	// iterate over all found relationships, and for each, iterate over all received relationships.
+	// if we can find the corresponding relationship, and their state match, continue.
+	// if their state differ or if we cannot find a match, append an update request
+	response := make(common.FederationState)
+	for _, r := range current {
```
maxlambrecht, line 156: This method needs to be refactored into smaller functions, there's too much nesting of control loops and if statements.

maxlambrecht (commented):
I have a big concern with the refactor of the server/endpoints: How would you do if you wanted later on to add a new api to one of the servers? Or if you wanted to expose one of the existing apis on a different transport (e.g. admin api on TCP)? You would need to refactor everything again, right?

pkg/server/api/harvester/handlers.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+)
+
+func (s *echoServer) syncBundleHandler(ctx echo.Context) error {
+	logger := ctx.Logger()
+
+	// auth
+	token := common.Member{
+		ID:          uuid.New(),
+		TrustDomain: spiffeid.RequireTrustDomainFromString("spiffe://td2"),
```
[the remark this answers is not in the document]
jufantozzi: Ops! this is a left over used for testing pourposes but I somehow missed it! thanks!
 I will update with the proper way to get the token

Makefile
```diff
@@ -140,6 +140,25 @@ test-unit:
 race-test:
 	go test -cover -race ./...
 
+
+## Temporary targets to test sync APIs
+create-relationship:
```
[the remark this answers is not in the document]
jufantozzi: my plan was to remove these soon in the future and have actual integration testing, WDYT?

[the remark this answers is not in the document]
maxlambrecht: I prefer to have separate from the main Makefile, but I leave it to you.

pkg/server/api/admin/handlers.go
```diff
@@ -0,0 +1,142 @@
...
+
+}
+
+func (s *localServer) createRelationshipHandler(w http.ResponseWriter, r *http.Request) {
+	logger := s.config.Logger
+
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		logger.Errorf("failed reading body: %v", err)
+		s.handleError(w, err)
```
mgbcaio, line 59: the `handleError` function already logs the error, do we need bot logs ? (handleError and the line 58) ?

pkg/server/api/harvester/handlers.go
```diff
@@ -0,0 +1,225 @@
...
+
+	"github.com/HewlettPackard/galadriel/pkg/common"
+	"github.com/HewlettPackard/galadriel/pkg/server/datastore"
+	"github.com/google/uuid"
+	"github.com/labstack/echo/v4"
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+)
+
+func (s *echoServer) syncBundleHandler(ctx echo.Context) error {
+	logger := ctx.Logger()
```
mgbcaio, line 19: Why aren't we using the directly the logger from the echoServer ? `s.config.Logger` Also applies to the other functions

jufantozzi:
>I have a big concern with the refactor of the server/endpoints: How would you do if you wanted later on to add a new api to one of the servers? Or if you wanted to expose one of the existing apis on a different transport (e.g. admin api on TCP)? You would need to refactor everything again, right?

Great point, yes, we would need more refactor to be able to have a way to properly "plug" different transports and be able to re-use the handlers. 

Im wondering if we split the handlers into their own packages, and have the run function receiving the transport type via config, it could work, right? If positive, I can try to implement it this way

pkg/server/api/harvester/handlers.go
jufantozzi: no real reason, we should be fixing that in #56

jufantozzi:
>>I have a big concern with the refactor of the server/endpoints: How would you do if you wanted later on to add a new api to one of the servers? Or if you wanted to expose one of the existing apis on a different transport (e.g. admin api on TCP)? You would need to refactor everything again, right?

>Great point @maxlambrecht, yes, we would need more refactor to be able to have a way to properly "plug" different transports and be able to re-use the handlers.
Im wondering if we split the handlers into their own packages, and have the run function receiving the transport type via config, it could work, right? If positive, I can try to implement it this way

Giving it a second thought, this goes back to some discussions we had a good time ago, where essentially these handlers would require library specific contracts (like echo.Context) and that would make it really difficult to be able to properly re-use it.

pkg/server/api/harvester/handlers.go
```diff
@@ -0,0 +1,225 @@
...
+	if !bytes.Equal(receivedHarvesterState.TrustBundleHash, currentState.TrustBundleHash) {
+		_, err := s.DataStore.UpdateMember(context.TODO(), token.TrustDomain.String(), &common.Member{
+			TrustBundle:     receivedHarvesterState.TrustBundle,
+			TrustBundleHash: receivedHarvesterState.TrustBundleHash,
+		})
+		if err != nil {
+			logger.Error("failed to update member: ", err)
+			return err
+		}
+		logger.Infof("Trust domain %s has been successfully updated", receivedHarvesterState.TrustDomain)
```
mchurichi, line 113: Same here. As a rule of thumb: use `Info` for high level component's lifecycle events and their relevant messages, and `Debug` for internal details that could be useful for someone looking deeper (bug hunting, performance issues, etc)
```suggestion
		logger.Debugf("Trust domain %s has been successfully updated", receivedHarvesterState.TrustDomain)
```

pkg/server/datastore/datastore.go
```diff
@@ -13,7 +13,12 @@ import (
 
 type DataStore interface {
 	CreateMember(ctx context.Context, m *common.Member) (*Member, error)
+	UpdateMember(ctx context.Context, trustDomain string, m *common.Member) (*Member, error)
+	GetMember(ctx context.Context, trustDomain string) (*Member, error)
```
mchurichi, line 17: What's the difference between `Get` and `Fetch`? if there is none, we should stick to one verb. I'd rather use Get but it's just a personal preference.

pkg/server/datastore/datastore.go
```diff
@@ -68,13 +75,50 @@ func (s *MemStore) CreateMember(_ context.Context, member *common.Member) (*Memb
...
+	return s.members[trustDomain], nil
+}
+
+func (s *MemStore) GetMember(_ context.Context, trustDomain string) (*Member, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	m, ok := s.members[trustDomain]
+	if !ok {
+		return nil, errors.New("failed updating member: member not found")
```
mchurichi, line 116:
```suggestion
		return nil, errors.New("failed getting member: member not found")
```

pkg/server/datastore/datastore.go
```diff
@@ -68,13 +75,50 @@ func (s *MemStore) CreateMember(_ context.Context, member *common.Member) (*Memb
...
+	if member.TrustBundleHash != nil {
+		s.members[trustDomain].TrustBundleHash = member.TrustBundleHash
+	}
+
+	return s.members[trustDomain], nil
+}
+
+func (s *MemStore) GetMember(_ context.Context, trustDomain string) (*Member, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
mchurichi, line 112:
```suggestion
	s.mu.RLock()
	defer s.mu.RUnlock()
```

pkg/server/datastore/datastore.go
```diff
@@ -100,6 +144,34 @@ func (s *MemStore) CreateRelationship(_ context.Context, rel *common.Relationshi
 	return r, nil
 }
 
+func (s *MemStore) GetRelationship(_ context.Context, trustDomain string) ([]*Relationship, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
mchurichi, line 149:
```suggestion
	s.mu.RLock()
	defer s.mu.RUnlock()
```

pkg/server/datastore/datastore.go
```diff
@@ -100,6 +144,34 @@ func (s *MemStore) CreateRelationship(_ context.Context, rel *common.Relationshi
 	return r, nil
 }
 
+func (s *MemStore) GetRelationship(_ context.Context, trustDomain string) ([]*Relationship, error) {
```
mchurichi, line 147:
```suggestion
func (s *MemStore) GetRelationships(_ context.Context, trustDomain string) ([]*Relationship, error) {
```

pkg/server/datastore/datastore.go
```diff
@@ -100,6 +144,34 @@ func (s *MemStore) CreateRelationship(_ context.Context, rel *common.Relationshi
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	var response []*Relationship
+
+	for _, r := range s.relationship {
+		if r.MemberA.TrustDomain == trustDomain || r.MemberB.TrustDomain == trustDomain {
+
+			memberA, ok := s.members[r.MemberA.TrustDomain]
+			if !ok {
+				return nil, errors.New("failed updating member: member not found")
```
mchurichi, line 157:
```suggestion
				return nil, errors.New("failed getting relationship: member not found")
```

pkg/server/datastore/datastore.go
```diff
@@ -100,6 +144,34 @@ func (s *MemStore) CreateRelationship(_ context.Context, rel *common.Relationshi
...
+			}
+			memberB, ok := s.members[r.MemberB.TrustDomain]
+			if !ok {
+				return nil, errors.New("failed updating member: member not found")
```
mchurichi, line 161:
```suggestion
				return nil, errors.New("failed getting member: member not found")
```

mchurichi (changes requested):
Thanks for putting this together! ❤️ I believe overall looks very good.
Just left some suggestions and questions here and there, mostly nitpicks about error logging, mutexes usage, and naming schemes. There are only a few breaking deals about bundle hash verification and authentication we need to take care of.

pkg/common/api.go
```diff
@@ -0,0 +1,38 @@
...
+
+// FederationState maps MemberState's by their trust domain.
+type FederationState map[string]MemberState
+
+// MemberState defines a member to be updated.
+// All returned MemberState's implicitly federates with the calling Harvester.
+type MemberState struct {
+	TrustDomain     string `json:"trust_domain"`
+	TrustBundle     []byte `json:"trust_bundle"`
+	TrustBundleHash []byte `json:"trust_bundle_hash"`
```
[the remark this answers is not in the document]
maxlambrecht: Discussed this offline with @mchurichi , crypto APIs accept and return `[]byte` type for hashes, so it makes sense to keep as that.

jufantozzi:
taking this one down in favor of splitting up the endpoints and the refactoring

PR #54, opened 2022-09-30
Send Trust Bundle from Harvester to Server

mchurichi:

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**
Adds the control loop task to watch for Bundle changes in the SPIRE Server and sends them to the Galadriel Server as bundle updates.

**Which issue this PR fixes**

### DIFF
--- a/cmd/harvester/cli/config.go
+++ b/cmd/harvester/cli/config.go
@@ -2,17 +2,20 @@ package cli
 
 import (
 	"fmt"
+	"io"
+	"time"
+
 	"github.com/HewlettPackard/galadriel/pkg/common/telemetry"
 	"github.com/HewlettPackard/galadriel/pkg/common/util"
 	"github.com/HewlettPackard/galadriel/pkg/harvester"
 	"github.com/hashicorp/hcl"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"io"
 )
 
 const (
-	defaultSpireSocketPath = "/tmp/spire-server/private/api.sock"
+	defaultSpireSocketPath       = "/tmp/spire-server/private/api.sock"
+	defaultBundleUpdatesInterval = "30s"
 )
 
 type Config struct {
@@ -20,8 +23,9 @@ type Config struct {
 }
 
 type harvesterConfig struct {
-	SpireSocketPath string `hcl:"spire_socket_path"`
-	ServerAddress   string `hcl:"server_address"`
+	SpireSocketPath       string `hcl:"spire_socket_path"`
+	ServerAddress         string `hcl:"server_address"`
+	BundleUpdatesInterval string `hcl:"bundle_updates_interval"`
 }
 
 // ParseConfig reads a configuration from the Reader and parses it
@@ -49,9 +53,14 @@ func NewHarvesterConfig(c *Config) (*harvester.Config, error) {
 		return nil, err
 	}
 
-	sc.SpireAddress = spireAddr
+	buInt, err := time.ParseDuration(c.Harvester.BundleUpdatesInterval)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse bundle updates interval: %v", err)
+	}
 
+	sc.SpireAddress = spireAddr
 	sc.ServerAddress = c.Harvester.ServerAddress
+	sc.BundleUpdatesInterval = buInt
 
 	sc.Log = logrus.WithField(telemetry.SubsystemName, telemetry.Harvester)
 
@@ -77,5 +86,9 @@ func newConfig(configBytes []byte) (*Config, error) {
 func (c *Config) setDefaults() {
 	if c.Harvester.SpireSocketPath == "" {
 		c.Harvester.SpireSocketPath = defaultSpireSocketPath
+	}
+
+	if c.Harvester.BundleUpdatesInterval == "" {
+		c.Harvester.BundleUpdatesInterval = defaultBundleUpdatesInterval
 	}
 }
--- a/pkg/harvester/client/server.go
+++ b/pkg/harvester/client/server.go
@@ -18,6 +18,7 @@ const (
 	contentType = "application/json"
 
 	postBundlePath = "/bundle"
+	onboardPath    = "/onboard"
 )
 
 // GaladrielServerClient represents a client to connect to Galadriel Server
@@ -37,14 +38,15 @@ type client struct {
 func NewGaladrielServerClient(address, token string) (GaladrielServerClient, error) {
 	return &client{
 		c:       *http.DefaultClient,
-		address: address,
+		address: "http://" + address,
 		token:   token,
 		logger:  logrus.WithField(telemetry.SubsystemName, telemetry.GaladrielServerClient),
 	}, nil
 }
 
 func (c *client) Connect(ctx context.Context, token string) error {
-	req, err := http.NewRequestWithContext(ctx, "CONNECT", fmt.Sprintf("http://%s/onboard", c.address), nil)
+	url := c.address + onboardPath
+	req, err := http.NewRequestWithContext(ctx, http.MethodConnect, url, nil)
 	if err != nil {
 		return err
 	}
--- a/pkg/harvester/client/server.go
+++ b/pkg/harvester/client/server.go
@@ -1,33 +1,44 @@
 package client
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"net/http"
 
+	"github.com/HewlettPackard/galadriel/pkg/common"
 	"github.com/HewlettPackard/galadriel/pkg/common/telemetry"
 	"github.com/sirupsen/logrus"
+)
+
+const (
+	contentType = "application/json"
+
+	postBundlePath = "/bundle"
 )
 
 // GaladrielServerClient represents a client to connect to Galadriel Server
 type GaladrielServerClient interface {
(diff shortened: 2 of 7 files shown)


pkg/harvester/client/server.go
```diff
@@ -3,24 +3,32 @@ package client
...
 
 type client struct {
+	c       *http.Client
 	address string
+	token   string
 	logger  logrus.FieldLogger
 }
 
-func NewGaladrielServerClient(address string) (GaladrielServerClient, error) {
+func NewGaladrielServerClient(address, token string) (GaladrielServerClient, error) {
```
maxlambrecht, line 27: I think it makes sense to pass the token here 👍

pkg/harvester/client/server.go
```diff
@@ -29,6 +37,15 @@ func (s *client) GetUpdates(ctx context.Context) ([]string, error) {
 	return nil, errors.New("not implemented")
 }
 
-func (s *client) PushUpdates(ctx context.Context, updates []string) error {
+func (s *client) PushBundle(ctx context.Context, req *common.PushBundleRequest) error {
```
maxlambrecht, line 40: Nit: would it make sense to use `Post` and `Get` in the the method names?

pkg/harvester/client/server.go
```diff
@@ -29,6 +37,15 @@ func (s *client) GetUpdates(ctx context.Context) ([]string, error) {
...
+	r, _ := http.NewRequestWithContext(ctx, "POST", fmt.Sprintf("http://%s/bundle", s.address), nil)
```
maxlambrecht, line 41: Nit: I guess we could store that formatted address of Galadriel server and just add the path here.

pkg/harvester/controller/controller.go
```diff
@@ -49,18 +55,49 @@ func (c *HarvesterController) Run(ctx context.Context) error {
...
+
+func (c *HarvesterController) hasBundleChanged(ctx context.Context, current *spiffebundle.Bundle) (*spiffebundle.Bundle, bool) {
+	b, err := c.spire.GetBundle(ctx)
+	if err != nil {
+		c.logger.Error("failed to check bundle updates")
+		return nil, false
+	}
+
+	if !current.Equal(b) {
+		c.logger.Infof("Authorities: %d", len(b.X509Bundle().X509Authorities()))
```
maxlambrecht, line 98: Remove this.

pkg/harvester/controller/controller.go
```diff
@@ -49,18 +55,49 @@ func (c *HarvesterController) Run(ctx context.Context) error {
...
+	})
+}
+
+func (c *HarvesterController) notifyBundleUpdates(ctx context.Context) error {
+	t := time.NewTicker(c.config.BundleUpdatesInterval)
+	var currentBundle *spiffebundle.Bundle
+
 	for {
 		select {
-		case <-ticker.C:
```
maxlambrecht, line 59: I think we should remove all these lines related to telemetry whenever we find them, right?

pkg/harvester/controller/controller.go
```diff
@@ -49,18 +55,49 @@ func (c *HarvesterController) Run(ctx context.Context) error {
...
 
 	<-ctx.Done()
+	c.logger.Debug("Done")
+
 	return nil
 }
 
 func (c *HarvesterController) run(ctx context.Context) {
-	ticker := time.NewTicker(2 * time.Second)
+	util.RunTasks(ctx, []func(context.Context) error{
```
maxlambrecht, line 64: Here it would be probably a good idea to handle the error returned by `RunTasks`

pkg/harvester/controller/controller.go
```diff
@@ -49,18 +55,49 @@ func (c *HarvesterController) Run(ctx context.Context) error {
...
-			return
+			return nil
 		}
 	}
 }
+
+func (c *HarvesterController) hasBundleChanged(ctx context.Context, current *spiffebundle.Bundle) (*spiffebundle.Bundle, bool) {
+	b, err := c.spire.GetBundle(ctx)
+	if err != nil {
+		c.logger.Error("failed to check bundle updates")
```
maxlambrecht, line 93: Log the `err` here.

maxlambrecht (commented):
It's looking good.

pkg/harvester/controller/controller.go
mchurichi: yeah, that makes sense, remove the noise for now

pkg/harvester/client/server.go
mchurichi: yeah, that sounds good

mgbcaio (approved):
LGTM 🤽

mgbcaio (approved):

jufantozzi (dismissed):
Neat, just one concern regarding storing an invalid state

pkg/harvester/client/server.go
```diff
@@ -1,33 +1,40 @@
...
+	token   string
 	logger  logrus.FieldLogger
 }
 
-func NewGaladrielServerClient(address string) (GaladrielServerClient, error) {
+func NewGaladrielServerClient(address, token string) (GaladrielServerClient, error) {
 	return &client{
 		c:       *http.DefaultClient,
-		address: address,
+		address: fmt.Sprintf("http://%s", address),
```
jufantozzi, line 36: since we're using the default client, we don't need to have the http url prefix

```suggestion
		address: fmt.Sprintf("%s", address),
```

cmd/harvester/cli/config.go
```diff
@@ -78,4 +86,8 @@ func (c *Config) setDefaults() {
 	if c.Harvester.SpireSocketPath == "" {
 		c.Harvester.SpireSocketPath = defaultSpireSocketPath
 	}
+
+	if c.Harvester.BundleUpdatesInterval == "" {
+		c.Harvester.BundleUpdatesInterval = "30s"
```
maxlambrecht, line 91: Nit: this magic number should be defined as a constant `defaultBundleUpdatesInterval`

maxlambrecht (approved):

pkg/harvester/controller/controller.go
```diff
@@ -48,18 +55,62 @@ func (c *HarvesterController) Run(ctx context.Context) error {
...
+				if err != nil {
+					c.logger.Error("failed to marshal X.509 bundle: %v", err)
+				}
+
+				err = c.server.PostBundle(ctx, &common.PostBundleRequest{
+					TrustDomain: b.TrustDomain().String(),
+					TrustBundle: x509b,
+				})
+				if err != nil {
+					c.logger.Errorf("failed to push X.509 bundle: %v", err)
```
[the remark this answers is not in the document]
mchurichi: good catch!

cmd/harvester/cli/config.go
mchurichi: definitely

pkg/harvester/controller/controller.go
```diff
@@ -48,18 +55,62 @@ func (c *HarvesterController) Run(ctx context.Context) error {
...
+func (c *HarvesterController) notifyBundleUpdates(ctx context.Context) error {
+	t := time.NewTicker(c.config.BundleUpdatesInterval)
+	var currentBundle *spiffebundle.Bundle
+
 	for {
 		select {
-		case <-ticker.C:
-			telemetry.Count(ctx, telemetry.HarvesterController, telemetry.TrustBundle, telemetry.Add)
+		case <-t.C:
+			b, changed := c.bundleHasChanged(ctx, currentBundle)
```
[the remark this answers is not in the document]
mchurichi: I went back and forth a few times with this one, IMO `hasBundleChanged` conforms to the _verb first_ naming scheme but reads a little awkward in this case,  `bundleHasChanged` in the other hand reads better but it does not conforms. I'd be open to alternatives too: `hasNewBundle` maybe? or something else

pkg/harvester/controller/controller.go
```diff
@@ -48,18 +55,62 @@ func (c *HarvesterController) Run(ctx context.Context) error {
 	go c.run(ctx)
 
 	<-ctx.Done()
+	c.logger.Debug("Done")
```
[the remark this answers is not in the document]
mchurichi: Right,`Done` says nothing to me, but the context thing sounds like an implementation detail that is not useful for a user looking at the logs. What about something like...
```suggestion
	c.logger.Debug("Shutting down...")
```

pkg/harvester/controller/controller.go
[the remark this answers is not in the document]
jufantozzi: I was trying to think something in the direction of `hasNewBundle` and I really like this one

pkg/harvester/controller/controller.go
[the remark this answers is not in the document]
jufantozzi: agreed!

pkg/harvester/client/server.go
```diff
@@ -1,39 +1,52 @@
...
+	token   string
 	logger  logrus.FieldLogger
 }
 
-func NewGaladrielServerClient(address string) (GaladrielServerClient, error) {
+func NewGaladrielServerClient(address, token string) (GaladrielServerClient, error) {
 	return &client{
 		c:       *http.DefaultClient,
-		address: address,
+		address: "http://" + address,
```
mchurichi, line 41: @jufantozzi the problem I believe you were facing was because here and in the Connect method we were prefixing the address with the http scheme. This should be ok now, tested with your branch and it works fine.

maxlambrecht (approved):
🚀

PR #56, opened 2022-10-03
Standardize logging and minor improvements

mchurichi:

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**
Standardizes loggers and removes Echo banners

**Which issue this PR fixes**

### DIFF
--- a/cmd/harvester/cli/cli.go
+++ b/cmd/harvester/cli/cli.go
@@ -11,12 +11,12 @@ var (
 	RootCmd      = NewRootCmd()
 	cmdExecute   = RootCmd.Execute
 	HarvesterCmd = &HarvesterCLI{
-		log: logrus.WithField(telemetry.SubsystemName, telemetry.Harvester),
+		logger: logrus.WithField(telemetry.SubsystemName, telemetry.Harvester),
 	}
 )
 
 type HarvesterCLI struct {
-	log logrus.FieldLogger
+	logger logrus.FieldLogger
 }
 
 func NewRootCmd() *cobra.Command {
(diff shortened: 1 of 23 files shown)


jufantozzi (commented):
I have only one comment, very nitpicky, but slightly _based_:
`Log/log > Logger/logger`

my arguments in favor are essentially:
1. this is a very personal one, but to me `logger.bla` reads better than `log.bla`;
2. many logging libraries, including `logrus` have their log object called "logger";
3. echo library has `.Logger()` method which we kinda have to use, to it would not be fully standarized.

mchurichi:
> I have only one comment, very nitpicky, but slightly _based_: `Log/log > Logger/logger`
> 
> my arguments in favor are essentially:
> 
>     1. this is a very personal one, but to me `logger.bla` reads better than `log.bla`;
> 
>     2. many logging libraries, including `logrus` have their log object called "logger";
> 
>     3. echo library has `.Logger()` method which we kinda have to use, to it would not be fully standarized.

Alright, I'm ok with either `log` or `logger` as long as we stick to one. I chose `log` just because it's shorter and easier to read but I will change it if you feel strongly about it. The only topic I disagree is the use of echo's Logger, as we should use our own to avoid logging formatting/configuration differences.

jufantozzi:
@mchurichi yup, that sounds good

jufantozzi (approved):
Thanks @mchurichi, this is great!

jufantozzi (approved):

mgbcaio (approved):
Thanks for that! 🍈

maxlambrecht (approved):
Nice! 🍰 

I prefer `logger` too.

mchurichi:
on hold until we get #52 merged

mchurichi:
This is now ready for review

maxlambrecht (approved):

PR #58, opened 2022-10-06
implements bundle endpoints

jufantozzi:
**Affected functionality**
Implements the two bundle endpoints `/bundle` and `/bundle/sync` 

Signed-off-by: Juliano Fantozzi <juliano.fantozzi@gmail.com>


jufantozzi:
We're getting a failed build from code duplication. This is mostly because there's a lot of copy paste required because we can't know if in the relationship, we're dealing with MemberA or MemberB. Refactoring for this is not worth since this behavior will change. 

As this should be solved once we have directional relationships, so I don't think it's worth to refactor this to fit the current code duplication threshold, I propose we temporarily increase our limit to avoid this refactor. (btw anyone knows the creds for that?)

Makefile-poc.mk
```diff
@@ -0,0 +1,25 @@
+## Temporary targets to test sync APIs
+create-relationship:
+	./bin/galadriel-server create member -t td1
+	./bin/galadriel-server create member -t td2
+	./bin/galadriel-server create relationship -a td1 -b td2
+
+.PHONY: create-relationship
+
+run-harvester: create-relationship
+	token=`./bin/galadriel-server generate token -t td2 | ggrep -Po "(?<=Access Token:\s).*"`; \
```
jufantozzi, line 10:
```suggestion
	token=`./bin/galadriel-server generate token -t td2 | grep -Po "(?<=Access Token:\s).*"`; \
```

dev/request_data/updated_bundle_sync.json
```diff
@@ -0,0 +1,13 @@
+{
```
jufantozzi, line 1: Delete this file

pkg/server/endpoints/management.go
```diff
@@ -199,6 +199,8 @@ func (e *Endpoints) validateToken(token string) (bool, error) {
 
 	e.Log.Debugf("Token valid for trust domain: %s\n", t.Member.TrustDomain)
 
+	c.Set("token", t)
```
mchurichi, line 202: nice!

pkg/server/datastore/datastore.go
```diff
@@ -78,6 +85,41 @@ func (s *MemStore) CreateMember(_ context.Context, member *common.Member) (*Memb
...
+	return s.members[trustDomain], nil
+}
+
+func (s *MemStore) GetMember(_ context.Context, trustDomain string) (*Member, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	m, ok := s.members[trustDomain]
+	if !ok {
+		return nil, errors.New("failed getting member: member not found")
```
mchurichi, line 117: same

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,243 @@
...
+	"errors"
+	"fmt"
+	"github.com/HewlettPackard/galadriel/pkg/common"
+	"github.com/HewlettPackard/galadriel/pkg/server/datastore"
+	"github.com/labstack/echo/v4"
+	"io"
+)
+
+func (e *Endpoints) syncBundleHandler(ctx echo.Context) error {
+	logger := ctx.Logger()
```
mchurichi, line 16: please use `e.Log`

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,243 @@
...
+	if !bytes.Equal(receivedHarvesterState.TrustBundleHash, currentState.TrustBundleHash) {
+		_, err := e.DataStore.UpdateMember(context.TODO(), token.Member.TrustDomain, &common.Member{
+			TrustBundle:     receivedHarvesterState.TrustBundle,
+			TrustBundleHash: receivedHarvesterState.TrustBundleHash,
+		})
+		if err != nil {
+			e.handleTcpError(ctx, fmt.Sprintf("failed to update member: %v", err))
+			return err
+		}
+		logger.Infof("Trust domain %s has been successfully updated", receivedHarvesterState.TrustDomain)
```
mchurichi, line 120:
```suggestion
		logger.Debugf("Trust domain %s has been successfully updated", receivedHarvesterState.TrustDomain)
```

pkg/server/endpoints/management.go
```diff
@@ -190,14 +190,16 @@ func (e *Endpoints) onboardHandler(c echo.Context) error {
 	return nil
 }
 
-func (e *Endpoints) validateToken(token string) (bool, error) {
-	t, err := e.DataStore.FetchAccessToken(context.TODO(), token)
+func (e *Endpoints) validateToken(c echo.Context, token string) (bool, error) {
```
mchurichi, line 193: nit: for consistency
```suggestion
func (e *Endpoints) validateToken(ctx echo.Context, token string) (bool, error) {
```

mchurichi (commented):
Thanks for this, this is starting to look good!
I'll do a second pass in the bundle state calculation shortly.

pkg/server/endpoints/harvester.go
jufantozzi: ops, forgot about this one

pkg/server/endpoints/harvester.go
jufantozzi: missed this one as well

pkg/server/datastore/datastore.go
```diff
@@ -14,11 +14,16 @@ import (
 
 type DataStore interface {
 	CreateMember(ctx context.Context, m *common.Member) (*Member, error)
+	UpdateMember(ctx context.Context, trustDomain string, m *common.Member) (*Member, error)
+	GetMember(ctx context.Context, trustDomain string) (*Member, error)
```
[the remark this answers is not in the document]
jufantozzi: Done

pkg/server/datastore/datastore.go
```diff
@@ -78,6 +85,41 @@ func (s *MemStore) CreateMember(_ context.Context, member *common.Member) (*Memb
...
+func (s *MemStore) UpdateMember(_ context.Context, trustDomain string, member *common.Member) (*Member, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if member == nil {
+		return nil, errors.New("query object cannot be nil")
+	}
+
+	if _, ok := s.members[trustDomain]; !ok {
+		return nil, errors.New("failed updating member: member not found")
```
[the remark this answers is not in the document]
jufantozzi: no, good catch

pkg/common/api.go
```diff
@@ -1,6 +1,36 @@
 package common
```
maxlambrecht, line 1: See https://github.com/HewlettPackard/galadriel/pull/62#discussion_r992396126

jufantozzi: do you agree this should come in a separate PR?

pkg/server/endpoints/run.go
```diff
@@ -122,4 +122,6 @@ func (e *Endpoints) addHandlers() {
 
 func (e *Endpoints) addTCPHandlers(server *echo.Echo) {
 	server.CONNECT("/onboard", e.onboardHandler)
+	server.POST("/bundle", e.postBundleHandler)
```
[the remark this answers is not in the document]
mchurichi: Even though it's not written in stone, plural resource names are usually considered to be _RESTful_, but we can't say we are strictly following REST rules here (URL should be something like `/members/<member_id>/bundles` in that case). That said, I'm ok with either plural or singular as long as we stay consistent.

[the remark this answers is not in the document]
jufantozzi: agreed with @mchurichi on this, specially since there will likely not be an actual plural action on any `/bundles` call

jufantozzi:
Ok this should be good for a second round @maxlambrecht @mchurichi

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,213 @@
...
+		return err
+	}
+
+	currentState, err := e.DataStore.GetMember(context.TODO(), token.TrustDomain.String())
+	if err != nil {
+		e.handleTcpError(ctx, err.Error())
+		return err
+	}
+
+	receivedHarvesterState.BundleDigest = util.GetDigest(receivedHarvesterState.Bundle)
```
mchurichi, line 105: as stated before, instead of overwriting whatever digest comes in the request, can we verify that it matches the calculated digest?

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	receivedHarvesterState := common.PostBundleRequest{}
+	err = json.Unmarshal(body, &receivedHarvesterState)
+	if err != nil {
+		e.handleTcpError(ctx, fmt.Sprintf("failed to unmarshal state: %v", err))
+		return err
+	}
+
+	currentState, err := e.DataStore.GetMember(context.TODO(), token.TrustDomain.String())
```
mchurichi, line 99: we should also double check that the trust domain from the request matches the one in the token

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,213 @@
...
+		if update != nil {
+			response[update.TrustDomain] = *update
+		}
+	}
+
+	return response
+}
+
+// findOutdatedRelationship looks for a federated entry found in r, in the received State.
+// If we find a match, we validate it's state and return the updated version.
```
mchurichi, line 154: nit:
```suggestion
// If we find a match, we validate its state and return the updated version.
```

pkg/harvester/controller/controller.go
```diff
@@ -87,8 +87,10 @@ func (c *HarvesterController) notifyBundleUpdates(ctx context.Context) error {
 				}
 
 				err = c.server.PostBundle(ctx, &common.PostBundleRequest{
-					TrustDomain: b.TrustDomain().String(),
-					TrustBundle: x509b,
+					TrustBundle: common.TrustBundle{
+						TrustDomain: b.TrustDomain(),
+						Bundle:      x509b,
```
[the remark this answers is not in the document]
jufantozzi: I actually went back and forth with this, ended up concluding that since we'd be signing it, it would be redundant, but if you have a strong opinion about it I can change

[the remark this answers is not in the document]
jufantozzi: (i also could be very much wrong about my conclusion xD)

pkg/server/endpoints/harvester.go
jufantozzi: Done

pkg/server/endpoints/run.go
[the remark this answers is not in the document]
maxlambrecht: well, `/bundles` is the resource, not the action, the action is the http verb GET, POST, etc. 

Nvm, I don't have a strong opinion in this regard.

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,223 @@
...
+	"errors"
+	"fmt"
+	"github.com/HewlettPackard/galadriel/pkg/common"
+	"github.com/HewlettPackard/galadriel/pkg/common/util"
+	"github.com/labstack/echo/v4"
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"io"
+)
+
+func (e *Endpoints) syncBundleHandler(ctx echo.Context) error {
```
maxlambrecht, line 16:
```suggestion
func (e *Endpoints) syncFederatedBundlesHandler(ctx echo.Context) error {
```

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,223 @@
...
+	"errors"
+	"fmt"
+	"github.com/HewlettPackard/galadriel/pkg/common"
+	"github.com/HewlettPackard/galadriel/pkg/common/util"
+	"github.com/labstack/echo/v4"
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"io"
+)
+
+func (e *Endpoints) syncFederatedBundleHandler(ctx echo.Context) error {
```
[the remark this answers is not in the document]
jufantozzi: image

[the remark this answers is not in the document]
jufantozzi: This is neat! Thanks for the suggestion!

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func getFederatedMembers(relationships []*common.Relationship, td spiffeid.TrustDomain) []*common.Member {
+	federatedMembers := make([]*common.Member, 0)
+
+	for _, r := range relationships {
+		ma := r.MemberA
+		mb := r.MemberB
+
+		if td.Compare(ma.TrustDomain) != 0 && ma != nil {
```
maxlambrecht, line 85: `ma == nil` would be an application error, it's not a normal condition, it should never happen that a relationship
is stored with a nil member, thus I'd rather not complicate the code here checking that condition. 

If you suspected that could happen, it would be preferable to check it early:
```
if ma == nil || mb == nil {
    return errors.New("invalid relationship found in datastore: member is nil")
}
```

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,189 @@
...
+func getFederatedBundlesUpdates(harvesterBundlesDigests common.BundlesDigests, federatedMembers []*common.Member) common.BundleUpdates {
+
+	response := make(common.BundleUpdates)
+
+	for _, m := range federatedMembers {
+		td := m.TrustDomain
+		serverDigest := m.BundleDigest
+		harvesterDigest := harvesterBundlesDigests[td]
+
+		// if the bundle digest the harvester has for a federated trust domain is not the same as the digest the server has,
```
mgbcaio, line 103: super nit: I think its more clear like this, took me a couple of minutes to understand the phrase xD. Feel free to ignore it.
⬇️ 
```suggestion
		// If the harvester has a bundle digest for a federated trust domain different from the the one the Server has,
```

pkg/server/endpoints/harvester.go
jufantozzi: We were getting a panic somewhere I can't really remember after I applied your suggestion, but the fix was to actually create the array we'r returning here with len(0) instead of len(relationships). 

My assumption is that when some member bundle does not yet exists (member was created but didn't connect yet) this function ends up with an array bigger than it actually populated, resulting in nil entries. 

This check was something I was trying to fix that didn't actually work but I originally thought It would be good to leave. I can't really disagree with you, though. That should be ensured by the application, layers above. I've removed the check.

pkg/server/endpoints/harvester.go
jufantozzi: I've tried improving the whole thing, PTAL.

pkg/server/endpoints/harvester.go
```diff
@@ -0,0 +1,205 @@
...
+		return err
+	}
+
+	federatedMembers := getFederatedMembers(relationships, harvesterTrustDomain)
+	lastBundlesDigests := getCurrentFederatedBundleDigests(federatedMembers)
+	bundlesUpdates := getFederatedBundlesUpdates(harvesterBundleDigests, federatedMembers)
+
+	response := common.SyncBundleResponse{}
+	response.State = lastBundlesDigests
+	response.Updates = bundlesUpdates
```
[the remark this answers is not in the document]
jufantozzi: nice catches

maxlambrecht (approved):
🚀

PR #59, opened 2022-10-07
Creates UDS file path if it doesn't exists already

mchurichi:

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
The Galadriel Server now automatically creates the full directory path to the UDS path configured where it will be served, and it won't fail if it doesn't exists already.

**Description of change**

**Which issue this PR fixes**

I will rebase and mark this a ready for review as soon as #58 is merged.

### DIFF
--- a/pkg/server/endpoints/run.go
+++ b/pkg/server/endpoints/run.go
@@ -5,12 +5,19 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
+	"path/filepath"
 
 	"github.com/HewlettPackard/galadriel/pkg/common/util"
 	"github.com/HewlettPackard/galadriel/pkg/server/datastore"
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"github.com/sirupsen/logrus"
+)
+
+const (
+	// owner has full access over the path, others can read and list
+	udsDirPermissions = 0755
 )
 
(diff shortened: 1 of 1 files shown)


maxlambrecht:
I thought this is already taken care of here: https://github.com/HewlettPackard/galadriel/blob/182a2309f8e254ee3cee699d780ecfbd0a50762b/pkg/server/endpoints/run.go#L31

Is it not?

pkg/server/endpoints/run.go
```diff
@@ -13,6 +15,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	// owner has full access over the path, others can read and list
+	udsDirPermissions = 0755
```
maxlambrecht, line 20: This could be `0770` or `0750` ... probably the latter is best

mchurichi:
> I thought this is already taken care of here:
> 
> https://github.com/HewlettPackard/galadriel/blob/182a2309f8e254ee3cee699d780ecfbd0a50762b/pkg/server/endpoints/run.go#L31
> 
> Is it not?

Oh you're right. I thought I ran into this issue but it was something else. Sometimes when the server is forced to exit (i.e. when you force-close the terminal where it's running) the UDS socket is not properly cleaned-up and the UDS file remains there, causing issues when you try to run it again. But this is a completely different thing, we can take care of that one later on.

I'll close this, thanks for the heads up.

PR #64, opened 2022-10-19
Cleanup all unused code and files

maxlambrecht:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
N/A

**Description of change**
Removing code and files that we are not using in the PoC and that won't probably be used afterwards.

**Which issue this PR fixes**


jufantozzi (approved):
😍

mgbcaio:
Any specific reason we are excluding the telemetry/prometheus folder? 
In the MVP backlog we are tracking those metrics to be part of Galadriel.

datastore.patch
```diff
@@ -0,0 +1,347 @@
+Index: pkg/common/api.go
```
jufantozzi, line 1: why add this to this PR?

maxlambrecht: committed by mistake

mchurichi (approved):
🍙

pkg/harvester/endpoints/config.go
```diff
@@ -15,8 +14,6 @@ type Config struct {
 	// LocalAddr is the local address to bind the listener to.
 	LocalAddress net.Addr
 
-	// Plugin catalog
-	Catalog catalog.Catalog
-
+	Log logrus.FieldLogger
```
mgbcaio, line 17: We already have the `Logger logrus.FieldLogger`, what is the purpose of adding this field ?

mchurichi (approved):

PR #66, opened 2022-10-24
adds support for log level

jufantozzi:
Signed-off-by: Juliano Fantozzi <juliano.fantozzi@hpe.com>

### DIFF
--- a/cmd/harvester/cli/config.go
+++ b/cmd/harvester/cli/config.go
@@ -26,6 +26,7 @@ type harvesterConfig struct {
 	SpireSocketPath       string `hcl:"spire_socket_path"`
 	ServerAddress         string `hcl:"server_address"`
 	BundleUpdatesInterval string `hcl:"bundle_updates_interval"`
+	LogLevel              string `hcl:"log_level"`
 }
 
 // ParseConfig reads a configuration from the Reader and parses it
@@ -89,5 +90,9 @@ func (c *Config) setDefaults() {
 
 	if c.Harvester.BundleUpdatesInterval == "" {
 		c.Harvester.BundleUpdatesInterval = defaultBundleUpdatesInterval
+	}
+
+	if c.Harvester.LogLevel == "" {
+		c.Harvester.LogLevel = "INFO"
 	}
 }
--- a/cmd/server/cli/config_test.go
+++ b/cmd/server/cli/config_test.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"errors"
 	"io"
+	"strings"
 	"testing"
 
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -20,6 +22,7 @@ func TestNewServerConfig(t *testing.T) {
 		ListenAddress: "localhost",
 		ListenPort:    8000,
 		SocketPath:    "/example",
+		LogLevel:      "INFO",
 	}}
 
 	sc, err := NewServerConfig(&config)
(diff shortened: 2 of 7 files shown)


cmd/harvester/cli/config.go
```diff
@@ -63,6 +64,16 @@ func NewHarvesterConfig(c *Config) (*harvester.Config, error) {
 
 	hc.Logger = logrus.WithField(telemetry.SubsystemName, telemetry.Harvester)
 
+	if c.Harvester.LogLevel == "" {
+		c.Harvester.LogLevel = "info"
```
mchurichi, line 68: I don't have a strong argument for this, and it doesn't make any difference for logrus, but log levels are usually all uppercase. Do you think we can change it here and in all occurrences?

cmd/harvester/cli/config.go
```diff
@@ -63,6 +64,16 @@ func NewHarvesterConfig(c *Config) (*harvester.Config, error) {
...
+	}
+	logLevel, err := logrus.ParseLevel(c.Harvester.LogLevel)
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.SetLevel(logLevel)
```
mchurichi, line 75: Not sure this is the most convenient place to set the log level. It feels this methods should only return newly created objects. How about moving this to the cmd definition, just after the `LoadConfig` call?

conf/server/server.conf
```diff
@@ -9,4 +9,7 @@ server {
     # socket_path: Path to bind the Galadriel Server API socket to.
     # Default: /tmp/galadriel-server/api.sock.
     socket_path = "/tmp/galadriel-server/api.sock"
-}
\ No newline at end of file
+
+    # log_level: Application log level.
+    #log_level = "info"
```
mchurichi, line 14:
```suggestion
    # log_level: Application log level. One of: TRACE, DEBUG, INFO, WARN, ERROR, FATAL, PANIC
    # Default: INFO
    log_level = "INFO"
```

cmd/harvester/cli/config.go
jufantozzi: We can't have this after the LoadConfig call because this config value does not return from this `LoadConfig` (the logger object gets created). We could add the log level to the returned config structures but not worth IMO. WDYT?

mgbcaio:
Other than what Maxi said, I got nothing for you, Juliano =) 🧮

cmd/harvester/cli/config.go
mchurichi: okay, I like it better where it is now 💯

jufantozzi:
> There are still some lowercase info though.

@mgbcaio the only one I could find left is for an automated test, does that also applies?

cmd/server/cli/config_test.go
```diff
@@ -20,6 +21,7 @@ func TestNewServerConfig(t *testing.T) {
 		ListenAddress: "localhost",
 		ListenPort:    8000,
 		SocketPath:    "/example",
+		LogLevel:      "info",
```
mchurichi, line 24:
```suggestion
		LogLevel:      "INFO",
```

mchurichi: Oh ok I see what you mean. In that case, I'd rather have this string match what we actually put there in a real scenario, that is all-uppercased log level, and lowercase it at the comparison level to match what GetLevel() outputs. We shouldn't be adapting our code to the dependencies, but the other way around.
See below.

cmd/server/cli/config_test.go
```diff
@@ -30,6 +32,7 @@ func TestNewServerConfig(t *testing.T) {
 	assert.Equal(t, "127.0.0.1", sc.TCPAddress.IP.String())
 	assert.Equal(t, config.Server.ListenPort, sc.TCPAddress.Port)
 	assert.Equal(t, config.Server.SocketPath, sc.LocalAddress.String())
+	assert.Equal(t, config.Server.LogLevel, logrus.GetLevel().String())
```
mchurichi, line 35:
```suggestion
	assert.Equal(t, strings.ToLower(config.Server.LogLevel), logrus.GetLevel().String())
```

mchurichi (approved):
:feijoada:

mgbcaio (dismissed):
🦧

mchurichi (approved):

mgbcaio (approved):

PR #70, opened 2022-12-06
[ATB-150] Open API 3.0 Specs - TCP & UDP - Victor

Victorblsilveira:
**Pull request checklist**

- [x] Documentation updated?

**Affected functionality**

No functionality really affected, only documentation API Specs

**Description of change**

Adding UDS and TCP OpenAPI 3.0 Specs

**Additional Stuff**
VS Code Plugin for OpenAPI 3.0 Specs Security Checks, visualization, and much more: 
- https://marketplace.visualstudio.com/items?itemName=42Crunch.vscode-openapi


pkg/server/openapi3/tcp_openapi3.yaml
```diff
@@ -0,0 +1,315 @@
...
+      properties:
+        state:
+          type: object
+          format: object
+          properties:
+            trust_domain: 
+              type: string
+              format: string
+            trust_bundle:
+              type: array
```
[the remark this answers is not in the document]
Victorblsilveira: It is totally possible, whatever we want. The current model was based on the POC data structures

pkg/server/openapi3/uds_openapi3.yaml
```diff
@@ -0,0 +1,629 @@
...
+        '200':
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Membership'
+      security:
+        - harvester_auth: []
+   
+  /tenant/{tenantID}/membership:
```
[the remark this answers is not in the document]
Victorblsilveira: Yes, is supposed to be the **PUT** at `/tenant/{tenantID}/membership/{membershipID}`

mchurichi:
Just to make sure I got APIs and responsibilities right, I'll try to recap and throw a few questions/doubts that came to my mind:
We have 4 personas here: Galadriel Admin (a.k.a. super user or box admin), Tenant Admin, Harvester, and Harvester Admin.

- Galadriel Admin
  - is the one who onboards new tenants
  - do we want to let him/her manage tenant resources? (CRUD operations over members, fed groups, etc)
- Tenant Admin
  - manages resources for a given tenant: onboards members, manage federation groups and requests memberships
  - do we want to let him/her manage harvester membership requests? (accept/deny requests)
- Harvester
  - takes care of all bundle exchange related tasks: pushes it's colocated spire server bundle, and sets foreign bundles
- Harvester Admin
  - is the one who -with the help of the Galadriel Admin- sets up the Harvester through the onboarding process
  - accepts/denies memberships requests

Given this, I wonder how we want to split our APIs. An API per persona? an API per transport protocol? We need to keep in mind that different methods (potentially in the same "API") may have different authorization mechanisms

Victorblsilveira:
> Just to make sure I got APIs and responsibilities right, I'll try to recap and throw a few questions/doubts that came to my mind: We have 4 personas here: Galadriel Admin (a.k.a. super user or box admin), Tenant Admin, Harvester, and Harvester Admin.
> 
> * Galadriel Admin
>   
>   * is the one who onboards new tenants
>   * do we want to let him/her manage tenant resources? (CRUD operations over members, fed groups, etc)
> * Tenant Admin
>   
>   * manages resources for a given tenant: onboards members, manage federation groups and requests memberships
>   * do we want to let him/her manage harvester membership requests? (accept/deny requests)
> * Harvester
>   
>   * takes care of all bundle exchange related tasks: pushes it's colocated spire server bundle, and sets foreign bundles
> * Harvester Admin
>   
>   * is the one who -with the help of the Galadriel Admin- sets up the Harvester through the onboarding process
>   * accepts/denies memberships requests
> 
> Given this, I wonder how we want to split our APIs. An API per persona? an API per transport protocol? We need to keep in mind that different methods (potentially in the same "API") may have different authorization mechanisms

I don't think that we should have APIs per persona since they can consume the same functionality. This is something that can be controlled through an authorization mechanism that we may consume. Initially, I would say to have one API for the TCP operations and another one for UDS. If we understand that is valid to split them ever further so we can do it, but for now I am not seeing a strong reason for that.

pkg/server/openapi3/tcp_openapi3.yaml
```diff
@@ -0,0 +1,315 @@
...
+              schema:
+                $ref: '#/components/schemas/Membership'
+      security:
+        - harvester_auth: []
+  
+  /tenant/{tenantID}/membership-request:
+    put:
+      tags:
+        - Membership
+      summary: Create a membership request
```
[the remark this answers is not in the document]
Victorblsilveira: Why not ? I mean, if a harvester has the access_token. Why did it cant request to join and Federation Group?

pkg/server/openapi3/tcp_openapi3.yaml
```diff
@@ -0,0 +1,315 @@
...
+        '200':
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Membership'
+      security:
+        - harvester_auth: []
+  
+  /tenant/{tenantID}/membership-request:
```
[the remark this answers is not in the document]
mchurichi: No, the approval/deny must be done in the Harvester side. This is how we enforce **mutual consent**: one side requests and the other approves, there's always two sides that reach an agreement and both must explicitly express their desire to establish the membership.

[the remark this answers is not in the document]
Victorblsilveira: Would we decide to use the implicit flow for now? This means that we will only need to accept the request from a member to join the Federation Group.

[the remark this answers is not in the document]
Victorblsilveira: I think a messed up some concepts. I will review the flow here: https://docs.google.com/document/d/1raWmBwPy-8cRwoGr-fLd6MtZxycccoOKdz-iZrcNh64/edit#

pkg/server/openapi3/tcp_openapi3.yaml
```diff
@@ -0,0 +1,315 @@
...
+        status: 
+          $ref: '#/components/schemas/BundleDigests'
+    BundleSyncResponse:
+      type: object
+      properties:
+        updates:
+          $ref: '#/components/schemas/BundleDigests'
+        state:
+          $ref: '#/components/schemas/BundleDigests'
+    MemberOnboard:
```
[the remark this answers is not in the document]
dennisgove: I think as a Bearer token is appropriate.

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,1045 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Federation'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /tenant/{tenantID}/join-token:
```
maxlambrecht, line 727: In my understanding this operation should include a MemberID. Join tokens are generated associated to a Member, in order to make it possible for Galadriel Server to associate a new Harvester to a Member.

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,1045 @@
...
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /tenant/{tenantID}/harvester/{harvesterID}:
```
maxlambrecht, line 265: This operation doesn't make much sense to me, where does the `harvesterID` come from? Also I think the creation of a new Harvester in Galadriel Server will happen when a new harvester connects to Galadriel Server for the first time. 
Any ideas? @mchurichi

maxlambrecht (commented):
This is looking great, @Victorblsilveira !

I think we still need to put some thinking into which operations we need to accomplish the process of creating/registering a new Harvester. I left some thought about that and comments about other operations. 

Do you plan to add the operation for adding a new tenant?

Victorblsilveira:
> This is looking great, @Victorblsilveira !
> 
> I think we still need to put some thinking into which operations we need to accomplish the process of creating/registering a new Harvester. I left some thought about that and comments about other operations.
> 
> Do you plan to add the operation for adding a new tenant?

Yes, I will add more endpoints to represent tenant operations. Tenant CRUD

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,331 @@
...
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                   type: string
+                   format: string
+        '401':
```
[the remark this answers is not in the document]
Victorblsilveira: Exactly that. Another important aspect that I am still a bit confused about is if we want to define every error message response explicitly or just use the default one

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,331 @@
...
+          $ref: '#/components/responses/Unauthorized'
+      security:
+        - harvester_auth: []
+
+  /tenant/{tenantID}/member/{memberID}/bundle/sync:
+    put:
```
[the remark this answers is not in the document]
Victorblsilveira: Can we pass the body in GET operations?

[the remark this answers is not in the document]
maxlambrecht: that's a good point, well, we can but it's not very standard. It seems confusing having a `PUT` operations whose descriptions is `Get bundles` though. Maybe we could use a `POST` and describe the operation as `Synchronize bundles with Galadriel Server sending the bundles digests the Harvester has`.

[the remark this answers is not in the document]
Victorblsilveira: I like that, lets's change it

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,1045 @@
...
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /tenant/{tenantID}/members/{memberID}:
```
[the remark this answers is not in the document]
Victorblsilveira: because I still need to change the harvester API. I was improving this one first since is the biggest one and has more details.

pkg/server/openapi3/management_openapi3.yaml
Victorblsilveira: This is to allow operations for management at the UDS socket. It is possible to get the `harvesterID` from the list endpoint.

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,1045 @@
...
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /tenant/{tenantID}/harvester:
```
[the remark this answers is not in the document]
Victorblsilveira: Not yet. Any ideas?

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,331 @@
...
+            format: int64
+      requestBody:
+        description: Add/Update a new bundle for a specific harverster from a specific tenant
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/BundlePut'
+        required: true
+      responses:
+        '200':
```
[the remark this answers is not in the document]
Victorblsilveira: should we change the method to `patch`?

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,331 @@
...
+      parameters:
+        - name: tenantID
+          in: path
+          description: ID of the tenant where the bundle will be part of
+          required: true
+          schema:
+            type: integer
+            format: int64
+      requestBody:
+        description: All the current digests from all bundles that it has.
```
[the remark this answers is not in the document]
Victorblsilveira: This is thrash documentation. See if I fixed it in recent commits.

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,331 @@
...
+          schema:
+            type: string
+            format: int64
+        - name: memberID
+          in: path
+          description: ID of the member which the bundle represents
+          required: true
+          schema:
+            type: string
+            format: int64
```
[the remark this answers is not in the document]
Victorblsilveira: Okay lets use UUIDs.

wibarre (commented):
For the transparency log validation, we need to add the signed-trust-bundle and the cert of the public key generated by Galadriel server.

mchurichi:
Just to summarize the thoughts I shared at yesterday's meeting: I think we need to be thoughtful on the permissions we give to admins. Allow admins to perform all CRUD operations over all resources just won't work, and it will cause more headaches than benefits (i.e. enable a tenant admin to update harvesters info could cause inconsistency issues, and even if we want to let them update some specific properties, we'd need to establish a granular permission system).

I emphasize the idea of continuing to focus on a clear definition of the personas, and the actions they would have to perform. This will help us think on the domain and the problem, and only then, drive the API design as a consequence.

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,331 @@
...
+              $ref: '#/components/schemas/BundleSync'
+        required: true
+      responses:
+        '200':
+          description: Get all the updates and the current state
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BundleSyncResponse'
+        '401':
```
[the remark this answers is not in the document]
Victorblsilveira: That's the idea.

pkg/server/openapi3/harvester_openapi3.yaml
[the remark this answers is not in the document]
Victorblsilveira: In conclusion, in the OpenAPI Specs, we will only define the success cases and the expected error message pattern as the Default format. The behaviors related to when to return each status code will be done in the implementation details as you mention.

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,466 @@
...
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /tenants/{tenantID}/member/{memberID}/csr:
+    put:
+      tags:
+        - Certificates
+      summary: It will generate a CSR to be signed by an Harvester
+      description: This endpoints should be used for generation CSR request to be signed in the Harvester side after the leader election proccess
```
mchurichi, line 227:
```suggestion
      summary: Signs an X.509 Certificate on behalf of the Harvester
      description: Receives a CSR and returns an X.509 Certificate signed by Galadriel Server's CA
```

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,466 @@
...
+                type: array
+                items:
+                  $ref: '#/components/schemas/Membership'
+                maxItems: 500
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /tenants/{tenantID}/member/{memberID}/csr:
```
mchurichi, line 222: A CSR is what the Harvester sends, part of the Certificate issuance procces, and it should be a property of request body.

How about naming it something like...
```suggestion
  /tenants/{tenantID}/member/{memberID}/certificates:
```

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,466 @@
...
+    UUID:
+      type: string
+      format: uuid
+      maxLength: 40
+      pattern: ^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$
+      example: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+    CSR:
+      type: object
+      properties:
+        certificate:
```
mchurichi, line 447:
```suggestion
        certificateSigningRequest: 
```
or `csr`

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,466 @@
...
+      pattern: ^( *[A-Z0-9a-z] *.*)*$
+      maxLength: 65
+      example: f0456d7aed088e791e4610c3c2ad63afe46e2e777988fdbc9270f15ec9711b42
+    UUID:
+      type: string
+      format: uuid
+      maxLength: 40
+      pattern: ^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$
+      example: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+    CSR:
```
mchurichi, line 444:
```suggestion
    CertificateRequest:
```

pkg/server/openapi3/harvester_openapi3.yaml
```diff
@@ -0,0 +1,466 @@
...
+      type: object
+      properties:
+        certificate: 
+          type: string
+          format: bytes
+          pattern: ^( *[A-Z0-9a-z] *.*)*$
+          maxLength: 400
+          example: MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAsgMI1AyV8J8TKf4rNRXm gBfeYVqZsP53vRRIgAyZboPELuTSqTZqXl1W4zkKKAGJXU/nsllMW6bfAavCRDMs OPKyAp/8VnRqSWrF+Gyw2CVoNtG6htqcFT3EqIazcRewD/Jl96OpAscjj+HHcxX4 4tQj6Y4e0zHOs/vN+YvHNjL6oG6ePNW09sqZos5pRRbMvZVikax6d8AuPUPSwmLH 3HDdrGD/n/CclMxMnISRnQjFCW03GNeVQGvqIO180EUKnIbSzTF73CrCxAHG/wfF MY+mZabcDhrrY0pbaZn1yaqRI9Q4MhMLV+9CGiUF0DzDCnzzx5q7ngN1N5lOg5pO BQIDAQAB
+    CSRResponse:
```
mchurichi, line 453:
```suggestion
    CertificateResponse:
```

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
+openapi: 3.0.3
+info:
+  title: Galadriel Server Management API - OpenAPI 3.0
+  description: |-
+    This is the UDS Galadriel OpenAPI 3.0 specs.
```
mchurichi, line 5:
```suggestion
    Galadriel Server Management API specification.
```

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+          in: path
+          description: Tenant identifier
+          required: true
+          schema:
+            $ref: '#/components/schemas/UUID'
+        - name: status
+          in: query
+          schema:
+            type: string
+            enum: [ACTIVE, DISABLED]
```
mchurichi, line 96: can we add all states here?

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+            $ref: '#/components/schemas/UUID'
+        - name: membershipID
+          in: path
+          description: ID of the membership
+          required: true
+          schema:
+            $ref: '#/components/schemas/UUID'
+      responses:
+        '200':
+          description: Successful deleted
```
mchurichi, line 291: this seems to be a leftover

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+            $ref: '#/components/schemas/UUID'
+        - name: federationGroupID
+          in: path
+          description: ID of the federation where the bundle will be part of
+          required: true
+          schema:
+            $ref: '#/components/schemas/UUID'
+      responses:
+        '200':
+          description: Successful deleted
```
mchurichi, line 389: leftover?

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+          schema:
+            $ref: '#/components/schemas/UUID'
+        - name: federationGroupID
+          in: path
+          description: ID of the federation where the bundle will be part of
+          required: true
+          schema:
+            $ref: '#/components/schemas/UUID'
+      requestBody:
+        description: Creates a new Federation Group
```
mchurichi, line 416:
```suggestion
        description: Updates a new Federation Group
```

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+          maxLength: 20
+          pattern: ^( *[A-Z0-9a-z] *)*$
+          example: A Federation name 
+        description:
+          type: string
+          format: string
+          maxLength: 200
+          pattern: ^( *[A-Z0-9a-z] *)*$
+          example: A good description about that federation group
+    Federation:
```
mchurichi, line 566:
```suggestion
    FederationGroup:
```

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+      type: object
+      additionalProperties: false
+      properties:
+        id: 
+          type: integer
+          format: int64
```
mchurichi, line 572: isn't this a UUID?

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+          format: date-time
+          maxLength: 21
+          example: "2021-01-30T08:30:00Z"
+    Member:
+      type: object
+      additionalProperties: false
+      properties:
+        id:
+         type: integer
+         format: int64
```
mchurichi, line 597: uuid

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+        tenantID: 
+         type: integer
+         format: int64
```
mchurichi, line 600: uuid

pkg/server/openapi3/management_openapi3.yaml
```diff
@@ -0,0 +1,706 @@
...
+      properties:
+        id:
+          type: integer
+          format: int64
+        tenantID:
+          type: integer
+          format: int64
+        memberID:
+          type: integer
+          format: int64
```
mchurichi, line 661: these should be UUIDs

pkg/server/openapi3/management_openapi3.yaml
maxlambrecht: All the IDs should be a UUID.

pkg/server/openapi3/harvester_openapi3.yaml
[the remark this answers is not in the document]
maxlambrecht: In this case, as we are updating the whole resource, we are storing a new bundle, `PUT` looks good to me.

pkg/server/openapi3/management_openapi3.yaml
Victorblsilveira: Yeah, have we defined all of them?

pkg/server/openapi3/harvester.yaml
```diff
@@ -0,0 +1,464 @@
...
+      additionalProperties: false
+      properties:
+        status: 
+          $ref: '#/components/schemas/BundleDigests'
+    BundleSyncResponse:
+      type: object
+      additionalProperties: false
+      properties:
+        updates:
+          $ref: '#/components/schemas/BundleDigests'
```
mchurichi, line 324: These should be Bundles, not the digest, as this is where the Harvester get bundles from

mchurichi: @jufantozzi do you recall if we have a doc detailing the requests and responses?

Victorblsilveira: @mchurichi  you mean both properties or just updates?

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BundleSyncResult'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /trust-domain/onboard:
```
jufantozzi, line 91: Isn't this path is to be used by the `management` (or `admin`) API?

jufantozzi (commented):
This is looking overall pretty good ⭐ great job @Victorblsilveira!

Most of comments are just minor fixes and trying to enforce consistency throughout the doc.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+          example: MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAsgMI1AyV8J8TKf4rNRXm gBfeYVqZsP53vRRIgAyZboPELuTSqTZqXl1W4zkKKAGJXU/nsllMW6bfAavCRDMs OPKyAp/8VnRqSWrF+Gyw2CVoNtG6htqcFT3EqIazcRewD/Jl96OpAscjj+HHcxX4 4tQj6Y4e0zHOs/vN+YvHNjL6oG6ePNW09sqZos5pRRbMvZVikax6d8AuPUPSwmLH 3HDdrGD/n/CclMxMnISRnQjFCW03GNeVQGvqIO180EUKnIbSzTF73CrCxAHG/wfF MY+mZabcDhrrY0pbaZn1yaqRI9Q4MhMLV+9CGiUF0DzDCnzzx5q7ngN1N5lOg5pO BQIDAQAB
+  securitySchemes:
+    harvester_auth:
+      type: "http"
+      scheme: "bearer"
+      bearerFormat: "JWT"
+    join_token:
+      type: "http"
+      scheme: "bearer"
+      bearerFormat: "JWT"
```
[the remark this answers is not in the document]
Victorblsilveira: From what I understand so far, yes.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+                items:
+                  $ref: '#/components/schemas/Relationship'
+                maxItems: 500
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /trust-domain/{trustDomainID}/certificates:
+    ## TODO move to the GCA component API
```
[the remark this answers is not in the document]
Victorblsilveira: Don't sure, best to do it's possible to remove it.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+    description: Operations related to relationship between trust domains
+  - name: Certificates
+    description: Operations related to certificates signing and public key sharing
+paths:
+  /trust-domain/{trustDomainID}/bundles:
+    put:
+      tags:
+        - Bundles
+      summary: Upload a new bundle to the server
+      description: Add/updates a trust bundle for a given trust domain and tenant
```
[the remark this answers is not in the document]
Victorblsilveira: I am removing tenants from the specs. If you see one, highlight it.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
+openapi: 3.0.3
+info:
+  title: Galadriel Server Harvester API - OpenAPI 3.0
```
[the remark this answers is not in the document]
Victorblsilveira: It is called harvester API due to the fact that is to serve the Harvesters. Feel free to suggest other names. DEMOCRACY

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+          example: spiffe://trust_domain/workload_identifier
+        onboarding_trust_bundle:
+          $ref: '#/components/schemas/TrustBundle'
+        description:
+          type: string
+          format: string
+          maxLength: 200
+          pattern: ^( *[A-Z0-9a-z] *)*$
+          example: ""
+    OnboardResult:
```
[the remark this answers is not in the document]
Victorblsilveira: We can simplify and only return the access_token. Dont sure, but if we follow the RESTFull principles we should return the entity being affected by that. Which means the entire trust domain.

pkg/server/api/openapi3/harvester/harvester.yaml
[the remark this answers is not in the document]
mchurichi: yes, let's remove this, we can add it later as part of the GCA API

pkg/server/api/openapi3/harvester/harvester.yaml
jufantozzi: you're right! My bad, I got confused here 😜

pkg/server/api/openapi3/harvester/harvester.yaml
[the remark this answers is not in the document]
Victorblsilveira: No, but in the onboarding process, the harvester will need to upload the onboarding trust_bundle and its SVID. From a restful perspective, ideally, the response should be the trust bundle entity with the properties updated. Or can be just the access token as you mention, it depends on what we are aiming for.

pkg/server/api/openapi3/harvester/harvester.yaml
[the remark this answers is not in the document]
maxlambrecht: The join_token will be a UUID like token, e.g. `d8ef05bd-1c2c-4b1d-b8e2-2ee40820047d`. And as the Onboard is a POST, it could go on the body.

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,314 @@
...
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+  version: 1.0.11
+servers:
+  - url: https://localhost.swagger.io/api/v3
+tags:
+  - name: Trust Domain
+    description: A representation of one or more SPIRE Servers that belong to the same Trust Domain.
+  - name: Relationships
+    description: A relationships is a representation of an association of a two different trust domains
```
maxlambrecht, line 19: "A relationship is the representation of a SPIFFE Federation Relationship between two Trust Domains"

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,314 @@
...
+      tags:
+        - Relationships
+      parameters:
+        - name: trustDomainID
+          in: query
+          description: TrustDomain ID
+          required: true
+          schema:
+            $ref: '#/components/schemas/UUID'
+        - name: status
```
maxlambrecht, line 84: This status needs to be reviewed. The relationship has two new fields, booleans, that tell who accepted the relationship: `trust_domain_a_consent` and `trust_domain_b_consent`.

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,314 @@
...
+    UUID:
+      type: string
+      format: uuid
+      maxLength: 40
+      pattern: ^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$
+      example: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+    SPIFEE_ID:
+      type: string
+      format: string
+      maxLength: 150
```
maxlambrecht, line 302: Review

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+    Onboard:
+      type: object
+      additionalProperties: false
+      properties:
+        trust_domain:
+          $ref: '#/components/schemas/TrustDomain'
+        harvester_spiffe_id:
+          type: string
+          format: string
+          maxLength: 150
```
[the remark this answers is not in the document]
maxlambrecht: yeah, it's an arbitrary restriction. I think it would make more sense to restrict based on the SPIFFE spec: 

```
URIs, as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986), do not have a maximal length. As an interoperability consideration, SPIFFE implementations MUST support SPIFFE URIs up to 2048 bytes in length and SHOULD NOT generate URIs of length greater than 2048 bytes.
```

So, 2048

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,314 @@
...
+          content:
+            application/json:
+              schema:
+                type: object
+                additionalProperties: false
+                properties:
+                  token:
+                   type: string
+                   maxLength: 200
+                   pattern: ^[A-Za-z0-9-_=]+\.[A-Za-z0-9-_=]+\.?[A-Za-z0-9-_.+/=]*$
```
maxlambrecht, line 173: The `join_token` is not a JWT token.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /trust-domain/{trustDomainID}/bundles/sync:
+    put:
+      tags:
+        - Bundles
+      summary: Get federated trust bundles updates from the Galadriel Server
+      description: Synchronize federated bundles with Galadriel Server
```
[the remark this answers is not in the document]
Victorblsilveira: +1 Trust Bundle

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+      format: string
+      maxLength: 2048
+      pattern: ^(.*[A-Za-z].)*$
+      example: "trust.domain.com"
+    TrustBundle:
+      type: string
+      format: bytes
+      pattern: ^( *[A-Z0-9a-z] *.*)*$
+      maxLength: 400
+      example: MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAsgMI1AyV8J8TKf4rNRXm gBfeYVqZsP53vRRIgAyZboPELuTSqTZqXl1W4zkKKAGJXU/nsllMW6bfAavCRDMs OPKyAp/8VnRqSWrF+Gyw2CVoNtG6htqcFT3EqIazcRewD/Jl96OpAscjj+HHcxX4 4tQj6Y4e0zHOs/vN+YvHNjL6oG6ePNW09sqZos5pRRbMvZVikax6d8AuPUPSwmLH 3HDdrGD/n/CclMxMnISRnQjFCW03GNeVQGvqIO180EUKnIbSzTF73CrCxAHG/wfF MY+mZabcDhrrY0pbaZn1yaqRI9Q4MhMLV+9CGiUF0DzDCnzzx5q7ngN1N5lOg5pO BQIDAQAB
```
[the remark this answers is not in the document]
Victorblsilveira: Yes, it is.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+      example: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+    CSR:
+      type: object
+      additionalProperties: false
+      properties:
+        certificate: 
+          type: string
+          format: bytes
+          pattern: ^( *[A-Z0-9a-z] *.*)*$
+          maxLength: 400
```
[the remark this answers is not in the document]
Victorblsilveira: Left unlimited can leave us open to overflow attacks.

pkg/server/api/openapi3/harvester/harvester.yaml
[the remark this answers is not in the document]
Victorblsilveira: I think that returning only the access token is the simplest solution, as @jufantozzi  mentioned, but do you guys see any necessity of returning the entire entity alongside the token?

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+      operationId: Onboard
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Onboard'
+        required: true
+      responses:
+        '200':
+          description: Successful operation. Returns a new join token to be used for authenticating the Trust Domain
```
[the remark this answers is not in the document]
Victorblsilveira: See if it was fixed

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+          $ref: '#/components/schemas/TrustBundle'
+        bundle_digest:
+          $ref: '#/components/schemas/BundleDigest'
+    BundlePutResult:
+      type: object
+      additionalProperties: false
+      properties:
+        status:
+          type: string
+          enum: [ACTIVE, DISABLED]
```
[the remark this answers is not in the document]
Victorblsilveira: Is to show the current status of the Trust Domain.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+      additionalProperties: false
+      properties:
+        status: 
+          $ref: '#/components/schemas/BundleDigests'
+    BundleSyncResult:
+      type: object
+      additionalProperties: false
+      properties:
+        updates:
+          $ref: '#/components/schemas/BundleDigests'
```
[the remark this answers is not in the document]
Victorblsilveira: The digests are not enough?

maxlambrecht (commented):
Hey @Victorblsilveira, these are the entities that are currently defined in Galadriel with the latest changes in the data model, do you think we can have those entities defined in this spec using the same fields and names?

```
type TrustDomain struct {
	ID                uuid.NullUUID
	Name              spiffeid.TrustDomain `json:"name"`
	Description       string               `json:"description"`
	HarvesterSpiffeID spiffeid.ID          `json:"harvester_spiffe_id"`
	OnboardingBundle  []byte               `json:"onboarding_bundle"`
	CreatedAt         time.Time            `json:"created_at"`
	UpdatedAt         time.Time            `json:"updated_at"`
}

type Relationship struct {
	ID                  uuid.NullUUID
	TrustDomainAID      uuid.UUID            `json:"trust_domain_a_id"`
	TrustDomainBID      uuid.UUID            `json:"trust_domain_b_id"`
	TrustDomainAName    spiffeid.TrustDomain `json:"trust_domain_a_name"`
	TrustDomainBName    spiffeid.TrustDomain `json:"trust_domain_b_name"`
	TrustDomainAConsent bool                 `json:"trust_domain_a_consent"`
	TrustDomainBConsent bool                 `json:"trust_domain_b_consent"`
	CreatedAt           time.Time            `json:"created_at"`
	UpdatedAt           time.Time            `json:"updated_at"`
}

type JoinToken struct {
	ID              uuid.NullUUID
	Token           string
	Used            bool
	TrustDomainID   uuid.UUID            `json:"trust_domain_id"`
	TrustDomainName spiffeid.TrustDomain `json:"trust_domain_name"`
	ExpiresAt       time.Time            `json:"expires_at"`
	CreatedAt       time.Time            `json:"created_at"`
	UpdatedAt       time.Time            `json:"updated_at"`
}

// Bundle represents a SPIFFE Trust bundle along with its digest.
type Bundle struct {
	ID                 uuid.NullUUID
	Data               []byte               `json:"bundle"`
	Digest             []byte               `json:"bundle_digest"`
	Signature          []byte               `json:"signature"`
	DigestAlgorithm    string               `json:"digest_algorithm"`
	SignatureAlgorithm string               `json:"signature_algorithm"`
	SigningCert        []byte               `json:"signing_cert"`
	TrustDomainID      uuid.UUID            `json:"trust_domain_id"`
	TrustDomainName    spiffeid.TrustDomain `json:"trust_domain_name"`
	CreatedAt          time.Time            `json:"created_at"`
	UpdatedAt          time.Time            `json:"updated_at"`
}
```

pkg/server/api/openapi3/harvester/harvester.yaml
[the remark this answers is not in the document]
Victorblsilveira: image

Victorblsilveira:
> with the latest changes i

That is a great idea. Not sure if will be possible to have a 1:1 mapping in the open API spec due to some limitations in the specification but it is worth trying.

pkg/server/api/openapi3/harvester/harvester.yaml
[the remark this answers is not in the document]
Victorblsilveira: We decide to remove the status and return no body for the bundle put response

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,313 @@
...
+        harvester_spiffe_id:
+          $ref: '#/components/schemas/SPIFEE_ID'
+        onboarding_trust_bundle:
+          $ref: '#/components/schemas/TrustBundle'
+        created_at:
+          type: string
+          format: date-time
+          maxLength: 21
+          example: "2021-01-30T08:30:00Z"
+        update_at:
```
wibarre, line 275: should this be updated_at? Matching our data model.

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,313 @@
...
+                $ref: '#/components/schemas/Relationship'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /relationships/{relationshipID}:
+    get:
+      tags:
+        - Relationships
```
wibarre, line 130: For consistency: We should add the summary of each get and put all around. I see it in some operations, but not all.

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,313 @@
...
+  - url: https://localhost.swagger.io/api/v3
+tags:
+  - name: Trust Domain
+    description: A SPIFFE Trust Domain
+  - name: Relationships
+    description: A relationship is the representation of a SPIFFE Federation Relationship between two Trust Domains
+  - name: Join Token
+    description:  Representation of a join token bound to a Trust Domain.
+    
+paths:
```
wibarre, line 23: Would it be beneficial to have a get by name?

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,313 @@
...
+                $ref: '#/components/schemas/TrustDomain'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /relationships:
+    get:
+      tags:
+        - Relationships
```
wibarre, line 76: Would this GET return all relationships in which trust domain (input parameter) is regardless if it is trust domain 1 or 2?  Do we want to distinguish between them?

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,358 @@
...
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /relationships:
+    get:
+      tags:
+        - Relationships
+      summary: Listing the relationships requests based on the query params.
```
wibarre, line 145: Is the intent of this GET to get all of the relationships a "harvester" has consented to? or all pending relationships? Or this a get based on a relationship id?  Should a harvester be able to query any relationship by id?  or should it be based on the trust_domain id?

wibarre (commented):
I focused mainly in the management.yaml. The mechanics of the comms between harvester and server is being worked on, so I am waiting to hear what the outcome is from that front.

pkg/server/api/openapi3/management/management.yaml
```diff
@@ -0,0 +1,314 @@
...
+    SPIFEE_ID:
+      type: string
+      format: string
+      maxLength: 150
+      example: spiffe://trust_domain/workload_identifier
+    TrustBundle:
+      type: string
+      format: bytes
+      pattern: ^( *[A-Z0-9a-z] *.*)*$
+      maxLength: 400
```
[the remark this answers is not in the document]
Victorblsilveira: I was taking a look at the open API specification and maxLength is related to string length rather than size in bytes. 
For restrictions in bytes size, we may need implementation reinforcement.

pkg/server/api/openapi3/management/management.yaml
Victorblsilveira: Do you mean query params for filtering by trust domain name?

pkg/server/api/openapi3/harvester/harvester.yaml
Victorblsilveira: Good questions, the endpoint can do almost all of this. It depends on the needs of our harvester since this API is for serving it. I see with good eyes getting the pending relationships for a given trust domain ID. This will facilitate Harvester Admin to accept/deny open relationship requests. Therefore the other ones are good to understand the system state, but I don't have in mind a strong use case for Glasgow POC

pkg/server/api/openapi3/management/management.yaml
wibarre: Yes. From the administrator stand point it would be handy the search by name rather than trust domain id.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,358 @@
...
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /trust-domain/onboard:
+    post:
+      tags:
+        - Onboard
+      summary: Onboarding a new Trust Domain in the Galadriel Server
+      description: Registers and assigns an identifier to a Trust Domain.
```
mchurichi, line 91: doesn't this endpoint belong to the admin API?

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,358 @@
...
+        status:
+          type: object
+          additionalProperties: true
+          properties:
+            trust_domain_1: 
+              $ref: '#/components/schemas/BundleDigest'
+          example:
+            trust_domain_1: f0456d7aed088e791e4610c3c2ad63afe46e2e777988fdbc9270f15ec9711b42
+            trust_domain_2: f0456d7aed088e791e4610c3c2ad63afe46e2e777988fdbc9270f15ec9711b42
+            trust_domain_n: f0456d7aed088e791e4610c3c2ad63afe46e2e777988fdbc9270f15ec9711b42
```
mchurichi, line 241: nit: can we have different digests here? 3 different valid SHA-256 hashes should work

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,358 @@
...
+      example: f0456d7aed088e791e4610c3c2ad63afe46e2e777988fdbc9270f15ec9711b42
+    UUID:
+      type: string
+      format: uuid
+      maxLength: 40
+      pattern: ^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$
+      example: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+    JWT:
+      type: string
+      format: uuid
```
mchurichi, line 336: JWTs are not UUIDs

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,358 @@
...
+      maxLength: 4096 # Number of characters
```
mchurichi, line 337: where does this max length come from?

pkg/server/api/openapi3/management/management.yaml
mchurichi: +1 (for all `{trustDomainID}` endpoints)

pkg/server/api/openapi3/management/management.yaml
mchurichi: Since relationships are bi-directional we don't need to distinguish between them

pkg/server/api/openapi3/harvester/harvester.yaml
Victorblsilveira: We don't, it's just a leftover.

pkg/server/api/openapi3/harvester/harvester.yaml
```diff
@@ -0,0 +1,394 @@
...
+      type: string
+      format: bytes
+      pattern: ^( *[A-Z0-9a-z] *.*)*$
+      maxLength: 400
+      example: MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAsgMI1AyV8J8TKf4rNRXm gBfeYVqZsP53vRRIgAyZboPELuTSqTZqXl1W4zkKKAGJXU/nsllMW6bfAavCRDMs OPKyAp/8VnRqSWrF+Gyw2CVoNtG6htqcFT3EqIazcRewD/Jl96OpAscjj+HHcxX4 4tQj6Y4e0zHOs/vN+YvHNjL6oG6ePNW09sqZos5pRRbMvZVikax6d8AuPUPSwmLH 3HDdrGD/n/CclMxMnISRnQjFCW03GNeVQGvqIO180EUKnIbSzTF73CrCxAHG/wfF MY+mZabcDhrrY0pbaZn1yaqRI9Q4MhMLV+9CGiUF0DzDCnzzx5q7ngN1N5lOg5pO BQIDAQAB
+    BundleDigest:
+      type: string
+      format: byte
+      pattern: ^( *[A-Z0-9a-z] *.*)*$
+      maxLength: 65
```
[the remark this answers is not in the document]
Victorblsilveira: I don't have sure that this length is in bytes. Looks like is more related to string size than anything else.

pkg/server/api/openapi3/harvester/harvester.yaml
Victorblsilveira: According to our documentation, the Onboard process is more related to the first contact between the Harvester and the GS/GCA components. The aspect of registering the Trust Domain is something else. 

Given that, I don't think so. The Admin API doesn't need an onboard endpoint. It needs an endpoint for registering the Trust Domain and another one for getting a Join Token for a Trust Domain. The onboard process will use later. 

Does that make sense?

pkg/server/api/openapi3/harvester/harvester.yaml
mchurichi: Oh okok, that makes sense, we should remove it then.

mchurichi: Actually, following what you suggested above, we can have an optional query param to filter by relationship status, as this is what the harvester will need, but at the same time an admin will need to know what are all of the relationships, being those pending, consented or non-consented.

pkg/server/api/openapi3/harvester/harvester.yaml
Victorblsilveira: JWT specification does not have a max length limit. Although usually, JWT tokens are stored in session cookies or HTTP header who has limits around 4096 and  8192 respectively. We can leave it open if you see that does make sense to limit it.

pkg/server/api/openapi3/harvester/harvester.yaml
Victorblsilveira: Relationship status is being removed. Do you see that it's needed? About the Admin needs, you are talking about the Harvester Admin, right? In that case, getting all relationships for a given Trust Domain, wouldn't supply what it needs to do so far? Or are you thinking of filters for each case?

pkg/server/api/openapi3/harvester/harvester.yaml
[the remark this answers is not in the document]
Victorblsilveira: I am using a visualizer in the VS Code. It is pretty similar to the swagger one.

pkg/server/api/openapi3/harvester/harvester.yaml
jufantozzi: +1 for using 8192

pkg/server/api/openapi3/harvester/harvester.yaml
Victorblsilveira: We add query params that will calculate the relationship status on the server side

pkg/server/api/management/management.yaml
jufantozzi, line 1: can we change the file names too? `management > admin`

Victorblsilveira: Sure, but let's wait a bit to see what the others think of it. I prefer management
+1 management

doc/API.md
```diff
@@ -0,0 +1,41 @@
...
+    
+    2.1 In [managment dir](../pkg/server/api/management/) run:
+    - `oapi-codegen -config managment.cfg.yaml management.yaml`
+
+
+## Generating code boilerplate using script
+
+The script automatize the process above.
+
+1. In [script dir](../scripts/) and just run:
```
mchurichi, line 40: do you think we can leverage the Makefile we already have and add a target there instead of creating more scripts?

Victorblsilveira: Sure, I think we can add the call to the script in the makefile or even delete the script and do that there. The only thing that concerns me, is that this makefile can become too big due to too many commands and responsibilities concentrated there. What do you think about this problem? Am I being too much cautious? Let me know what you think.

Makefile
```diff
@@ -199,12 +199,18 @@ help:
...
-generate: generatesql
+generate: generatesql generete-spec
 
 generatesql:
 	$(sqlc_bin) generate --file $(sqlc_config_file)
+
+generete-spec:
+	cd ./pkg/common/api; ${GOPATH}/bin/oapi-codegen -config schemas.cfg.yaml schemas.yaml
+	cd ./pkg/server/api/harvester; ${GOPATH}/bin/oapi-codegen -config harvester.cfg.yaml harvester.yaml
+	cd ./pkg/server/api/management;	${GOPATH}/bin/oapi-codegen -config management.cfg.yaml management.yaml
```
mchurichi, line 215: 40+ lines to 3 lines, love it! 💯

pkg/server/api/harvester/harvester.yaml
```diff
@@ -0,0 +1,284 @@
...
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /trust-domain/{trustDomainName}/bundles/sync:
+    post:
+      tags:
+        - Trust Bundles
+      summary: Get federated trust bundles updates from the Galadriel Server
+      description: Synchronize federated bundles with Galadriel Server
```
mchurichi, line 58:
```suggestion
      summary: Synchronizes federated bundles with Galadriel Server
```

pkg/server/api/harvester/harvester.yaml
```diff
@@ -0,0 +1,284 @@
...
+          $ref: '../../../common/api/schemas.yaml#/components/schemas/TrustBundle'
+        digest:
+          $ref: '../../../common/api/schemas.yaml#/components/schemas/BundleDigest'
+    BundleDigests:
+      type: object
+      additionalProperties: false
+      properties:
+        status:
+          type: object
+          additionalProperties: true
```
mchurichi, line 210: looks like what we want to define here is a [Dictionary](https://swagger.io/docs/specification/data-models/dictionaries/) without fixed keys

pkg/server/api/harvester/harvester.yaml
```diff
@@ -0,0 +1,284 @@
...
+          properties:
+            trust_domain_1:
```
mchurichi, line 212: this should go away

pkg/server/api/harvester/harvester.yaml
```diff
@@ -0,0 +1,284 @@
...
+      additionalProperties: false
+      properties:
+        updates:
+          type: array
+          items:
+            $ref: '#/components/schemas/TrustBundleSync'
+        state:
+          type: array
+          items:
+            $ref: '#/components/schemas/TrustBundleSync'
```
mchurichi, line 231: could you please sync with @jufantozzi on what should be in both `updates` and `state` here?

pkg/common/api/schemas.cfg.yaml
```diff
@@ -0,0 +1,9 @@
+package: common
+generate:
+  echo-server: false
+  client: false
+  models: true
+  embedded-spec: true
+output: schemas.gen.go
+output-options:
+  skip-prune: true
```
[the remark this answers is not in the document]
Victorblsilveira: All definitions in the documents should be used, otherwise, should be deleted it. So, no reason for pruning.

pkg/server/api/management/management.yaml
Victorblsilveira: sold to admin

pkg/server/api/harvester/harvester.yaml
```diff
@@ -0,0 +1,284 @@
...
+          maxItems: 100
+    Onboard:
+      type: object
+      additionalProperties: false
+      properties:
+        trust_domain:
+          $ref: '../../../common/api/schemas.yaml#/components/schemas/TrustDomainName'
+        harvester_spiffe_id:
```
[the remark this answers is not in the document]
Victorblsilveira: Join tokens are supposed to be UUID Tokens, can we store such kind of information there? 
@maxlambrecht mention [here](https://github.com/HewlettPackard/galadriel/pull/70#discussion_r1126862668) that is only supposed to send the trust bundle. So in that sense, the only thing that we need to send here is the trust_bundle and the description. Is that right?

pkg/server/api/harvester/harvester.yaml
```diff
@@ -0,0 +1,284 @@
...
+      properties:
+        trust_domain: 
+          $ref: '../../../common/api/schemas.yaml#/components/schemas/TrustDomainName'
+        trust_bundle:
+          $ref: '../../../common/api/schemas.yaml#/components/schemas/TrustBundle'
+        bundle_digest:
+          $ref: '../../../common/api/schemas.yaml#/components/schemas/BundleDigest'
+        signature:
+          $ref: '#/components/schemas/Signature'
+        signing_cert:
```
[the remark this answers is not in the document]
maxlambrecht: We can agree any of those options and I can change my models accordingly.

+1 signing_certificate

Also we aren't going to need the `bundle_digest`, I just realized it. I'm removing it from the models and the design documents.

mchurichi (commented):
Just a suggestion for the sake of the review process: could you please wait until pushing the changes to mark a comment as resolved? Sometimes, I just don't know if a comment went stale or was discarded, and sometimes I doubt whether or not I made the suggestion I meant to 😆

pkg/common/api/schemas.yaml
```diff
@@ -0,0 +1,179 @@
...
+      type: string
+      format: string
+      maxLength: 4096 # Number of characters
+      pattern: ^[a-zA-Z0-9\-_]+?\.[a-zA-Z0-9\-_]+?\.([a-zA-Z0-9\-_]+)?$
+      example: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
+    Certificate:
+      description: Base-64 encoded Trust Bundle in PEM format
+      type: string
+      format: byte
+      example: |+
```
mchurichi, line 123: Even when `format` is an open property, according to swagger documentation `byte` means `base64-encoded characters, for example, U3dhZ2dlciByb2Nrcw==`.

This can be a bit confusing since PEM is already base64-encoded + header and footer.

I'll suggest to:
- use a custom `pem` format that should help make it clear what's the content of this field (or even leave it out)
- use a fully formatted X.509 certificate in PEM format as an example, including the header and footer

We can have a custom schema for this field.

```yaml
type: string
format: pem
example: |
    -----BEGIN CERTIFICATE-----
    MIIH/TCCBeWgAwIBAgIQaBYE3/M08XHYCnNVmcFBcjANBgkqhkiG9w0BAQsFADBy
    MQswCQYDVQQGEwJVUzEOMAwGA1UECAwFVGV4YXMxEDAOBgNVBAcMB0hvdXN0b24x
    ETAPBgNVBAoMCFNTTCBDb3JwMS4wLAYDVQQDDCVTU0wuY29tIEVWIFNTTCBJbnRl
    cm1lZGlhdGUgQ0EgUlNBIFIzMB4XDTIwMDQwMTAwNTgzM1oXDTIxMDcxNjAwNTgz
    M1owgb0xCzAJBgNVBAYTAlVTMQ4wDAYDVQQIDAVUZXhhczEQMA4GA1UEBwwHSG91
    c3RvbjERMA8GA1UECgwIU1NMIENvcnAxFjAUBgNVBAUTDU5WMjAwODE2MTQyNDMx
    FDASBgNVBAMMC3d3dy5zc2wuY29tMR0wGwYDVQQPDBRQcml2YXRlIE9yZ2FuaXph
    dGlvbjEXMBUGCysGAQQBgjc8AgECDAZOZXZhZGExEzARBgsrBgEEAYI3PAIBAxMC
    VVMwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDHheRkbb1FCc7xRKst
    wK0JIGaKY8t7JbS2bQ2b6YIJDgnHuIYHqBrCUV79oelikkokRkFvcvpaKinFHDQH
    UpWEI6RUERYmSCg3O8Wi42uOcV2B5ZabmXCkwdxY5Ecl51BbM8UnGdoAGbdNmiRm
    SmTjcs+lhMxg4fFY6lBpiEVFiGUjGRR+61R67Lz6U4KJeLNcCm07QwFYKBmpi08g
    dygSvRdUw55Jopredj+VGtjUkB4hFT4GQX/ght69Rlqz/+8u0dEQkhuUuucrqalm
    SGy43HRwBfDKFwYeWM7CPMd5e/dO+t08t8PbjzVTTv5hQDCsEYIV2T7AFI9ScNxM
    kh7/AgMBAAGjggNBMIIDPTAfBgNVHSMEGDAWgBS/wVqH/yj6QT39t0/kHa+gYVgp
    vTB/BggrBgEFBQcBAQRzMHEwTQYIKwYBBQUHMAKGQWh0dHA6Ly93d3cuc3NsLmNv
    bS9yZXBvc2l0b3J5L1NTTGNvbS1TdWJDQS1FVi1TU0wtUlNBLTQwOTYtUjMuY3J0
    MCAGCCsGAQUFBzABhhRodHRwOi8vb2NzcHMuc3NsLmNvbTAfBgNVHREEGDAWggt3
    d3cuc3NsLmNvbYIHc3NsLmNvbTBfBgNVHSAEWDBWMAcGBWeBDAEBMA0GCyqEaAGG
    9ncCBQEBMDwGDCsGAQQBgqkwAQMBBDAsMCoGCCsGAQUFBwIBFh5odHRwczovL3d3
    dy5zc2wuY29tL3JlcG9zaXRvcnkwHQYDVR0lBBYwFAYIKwYBBQUHAwIGCCsGAQUF
    BwMBMEgGA1UdHwRBMD8wPaA7oDmGN2h0dHA6Ly9jcmxzLnNzbC5jb20vU1NMY29t
    LVN1YkNBLUVWLVNTTC1SU0EtNDA5Ni1SMy5jcmwwHQYDVR0OBBYEFADAFUIazw5r
    ZIHapnRxIUnpw+GLMA4GA1UdDwEB/wQEAwIFoDCCAX0GCisGAQQB1nkCBAIEggFt
    BIIBaQFnAHcA9lyUL9F3MCIUVBgIMJRWjuNNExkzv98MLyALzE7xZOMAAAFxM0ho
    bwAABAMASDBGAiEA6xeliNR8Gk/63pYdnS/vOx/CjptEMEv89WWh1/urWIECIQDy
    BreHU25DzwukQaRQjwW655ZLkqCnxbxQWRiOemj9JAB1AJQgvB6O1Y1siHMfgosi
    LA3R2k1ebE+UPWHbTi9YTaLCAAABcTNIaNwAAAQDAEYwRAIgGRE4wzabNRdD8kq/
    vFP3tQe2hm0x5nXulowh4Ibw3lkCIFYb/3lSDplS7AcR4r+XpWtEKSTFWJmNCRbc
    XJur2RGBAHUA7sCV7o1yZA+S48O5G8cSo2lqCXtLahoUOOZHssvtxfkAAAFxM0ho
    8wAABAMARjBEAiB6IvboWss3R4ItVwjebl7D3yoFaX0NDh2dWhhgwCxrHwIgCfq7
    ocMC5t+1ji5M5xaLmPC4I+WX3I/ARkWSyiO7IQcwDQYJKoZIhvcNAQELBQADggIB
    ACeuur4QnujqmguSrHU3mhf+cJodzTQNqo4tde+PD1/eFdYAELu8xF+0At7xJiPY
    i5RKwilyP56v+3iY2T9lw7S8TJ041VLhaIKp14MzSUzRyeoOAsJ7QADMClHKUDlH
    UU2pNuo88Y6igovT3bsnwJNiEQNqymSSYhktw0taduoqjqXn06gsVioWTVDXysd5
    qEx4t6sIgIcMm26YH1vJpCQEhKpc2y07gRkklBZRtMjThv4cXyyMX7uTcdT7AJBP
    ueifCoV25JxXuo8d5139gwP1BAe7IBVPx2u7KN/UyOXdZmwMf/TmFGwDdCfsyHf/
    ZsB2wLHozTYoAVmQ9FoU1JLgcVivqJ+vNlBhHXhlxMdN0j80R9Nz6EIglQjeK3O8
    I/cFGm/B8+42hOlCId9ZdtndJcRJVji0wD0qwevCafA9jJlHv/jsE+I9Uz6cpCyh
    sw+lrFdxUgqU58axqeK89FR+No4q0IIO+Ji1rJKr9nkSB0BqXozVnE1YB/KLvdIs
    uYZJuqb2pKku+zzT6gUwHUTZvBiNOtXL4Nxwc/KT7WzOSd2wP10QI8DKg4vfiNDs
    HWmB1c4Kji6gOgA5uSUzaGmq/v4VncK5Ur+n9LbfnfLc28J5ft/GotinMyDk3iar
    F10YlqcOmeX1uFmKbdi/XorGlkCoMF3TDx8rmp9DBiB/
    -----END CERTIFICATE-----
```

mchurichi: @maxlambrecht @jufantozzi @wibarre thoughts?

jufantozzi: +1

pkg/server/api/harvester/harvester.yaml
Victorblsilveira: I spoke with @jufantozzi (correct me if I am wrong) regarding this. The conclusion was, for now, the changes regarding if we are going to send the consent signature alongside the bundles, need a design discussion.

pkg/common/api/schemas.cfg.yaml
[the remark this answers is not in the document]
mchurichi: sounds good

pkg/server/api/harvester/harvester.yaml
mchurichi: Ok, that makes sense, but still, and AFAIK `updates` and `state` shouldn't have the same structure.
`state` holds the digests of ALL the bundles a Harvester knows about.
`updates` holds all the NEW bundles a Harvester needs to set in SPIRE.

With @jufantozzi we worked on a [flow breakdown doc](https://docs.google.com/document/d/1V9Tky5B0Udk3jmbHVjOTZcAloizvUlpGj14s3xsP_P0/edit?usp=sharing) a while ago, it may be a bit outdated though.

pkg/server/api/harvester/harvester.yaml
```diff
@@ -0,0 +1,274 @@
...
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /trust-domain/{trustDomainName}/bundles/sync:
+    post:
+      tags:
+        - Trust Bundles
+      summary: Synchronizes federated bundles with Galadriel Server
```
wibarre, line 56: Is the purpose of this to "pull" the trust bundles that were rotated and updated from the server? If yes, should this be a GET?

wibarre (commented):
Thank you Victor for the great work.

pkg/server/api/harvester/harvester.yaml
maxlambrecht: > `updates` and `state` shouldn't have the same structure. `state` holds the digests of ALL the bundles a Harvester knows about. `updates` holds all the NEW bundles a Harvester needs to set in SPIRE.
> 

This is correct.

maxlambrecht: > The conclusion was, for now, the changes regarding if we are going to send the consent signature alongside the bundles, need a design discussion.

This is another discussion, in the new bundles (so in `updates`) each bundle will come with a consent, but this is still TBD.

pkg/server/api/harvester/harvester.yaml
maxlambrecht: GET operations don't include a body, this one needs to post the Harvester "state"

pkg/server/api/harvester/harvester.yaml
jufantozzi: We missed that the object for responses were the same, they're not. One maps trust-domains to digests, and the other maps trust-domains to bundles. I lead to this confusion because I missed this and thought you're talking about something else @mchurichi 😓 my bad

pkg/common/api/schemas.yaml
maxlambrecht: I wonder why we don't use a binary format and transmit the bundles in ASN.1 DER encoded bytes, which is more efficient, as PEM is 33% larger being base64 encoded.

pkg/server/api/admin/admin.yaml
```diff
@@ -0,0 +1,244 @@
...
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - harvester_auth: []
+
+  /relationships:
+    get:
+      tags:
+        - Relationships
+      summary: Get relationships based on filters
```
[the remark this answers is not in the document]
Victorblsilveira: To do: 
- Use enumeration for status 
- Change the trustDomain -> trustDomainName.

pkg/common/api/schemas.yaml
Victorblsilveira: We will keep PEM for now and considering change it later. 
JIRA Ticket: https://hpe.atlassian.net/browse/GLCP-41060

doc/API.md
```diff
@@ -0,0 +1,38 @@
...
+    > `go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@latest`
+
+2. Go to where the specs are (listed above), and generate the boilerplate
+
+    > `oapi-codegen -config <config-file> <openapi_spec.yaml>`    
+
+    Example: 
+    
+    2.1 While in the [management](../pkg/server/api/management/) directory, run the following command:
+    - `oapi-codegen -config management.cfg.yaml management.yaml`
```
mchurichi, line 31: update to use `admin` instead of `management`

pkg/common/api/schemas.yaml
maxlambrecht: Ok, sounds good. We'd need to use [BSON](https://bsonspec.org/) instead of JSON to transmit the data.

pkg/common/api/schemas.yaml
```diff
@@ -0,0 +1,181 @@
...
+          maximum: 599
+          minimum: 100
+    TrustDomainName:
+      type: string
+      format: string
+      maxLength: 2048
+      example: "trust.domain.com"
+    BundleDigest:
+      type: string
+      format: byte
```
mchurichi, line 27: Same thing as trust bundle format.

The hashing algorithms can output a short set of characters we can easily handle as simple strings. If we want to store it as `byte`s (I don't see a reason to do so) we should have them base-64-encoded, and the example above needs to be changed accordingly.

I'll just remove the format for now.

pkg/common/api/schemas.yaml
```diff
@@ -0,0 +1,181 @@
...
+        ZsB2wLHozTYoAVmQ9FoU1JLgcVivqJ+vNlBhHXhlxMdN0j80R9Nz6EIglQjeK3O8
+        I/cFGm/B8+42hOlCId9ZdtndJcRJVji0wD0qwevCafA9jJlHv/jsE+I9Uz6cpCyh
+        sw+lrFdxUgqU58axqeK89FR+No4q0IIO+Ji1rJKr9nkSB0BqXozVnE1YB/KLvdIs
+        uYZJuqb2pKku+zzT6gUwHUTZvBiNOtXL4Nxwc/KT7WzOSd2wP10QI8DKg4vfiNDs
+        HWmB1c4Kji6gOgA5uSUzaGmq/v4VncK5Ur+n9LbfnfLc28J5ft/GotinMyDk3iar
+        F10YlqcOmeX1uFmKbdi/XorGlkCoMF3TDx8rmp9DBiB/
+        -----END CERTIFICATE-----
+    Signature:
+      type: string
+      format: byte
```
mchurichi, line 171: same

pkg/common/api/schemas.yaml
mchurichi: 🤝

pkg/common/api/schemas.yaml
```diff
@@ -0,0 +1,181 @@
...
+      pattern: ^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$
+      example: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+    JWT:
+      type: string
+      format: jwt
+      maxLength: 4096 # Number of characters
+      pattern: ^[a-zA-Z0-9\-_]+?\.[a-zA-Z0-9\-_]+?\.([a-zA-Z0-9\-_]+)?$
+      example: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
+    Certificate:
+      description: Base-64 encoded Trust Bundle in PEM format
```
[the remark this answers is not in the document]
mchurichi: actually, this is not a Trust Bundle anymore, disregard my previous comment
```suggestion
      description: X.509 certificate in PEM format
```

wibarre (dismissed):
Thank you Victor for retrofitting all the feedback into this PR.

maxlambrecht (approved):
Excellent work @Victorblsilveira!

mchurichi (approved):
🙌

PR #73, opened 2022-12-15
Add Datastore layer

maxlambrecht:

**Pull request check list**

- [X] Documentation updated?

**Affected functionality**
Added a new Datastore implementation that supports Postgres

**Description of change**

*  This PR adds migration SQLs to create the schema based on the new data model, and also adds the SQL Queries. 
* The interfaces, model structs, and lower level DB operations were auto generated using [sqlc](https://github.com/kyleconroy/sqlc) which uses the SQL files from `migrations` and `queries` as input.
* A new `Datastore` struct was added, that defines CRUD methods for all the entities of Galadriel.
* When the `Datastore` is created and it connects to the configured DB, the schema is validated and migrations applied if needed, using [golang-migrate](https://github.com/golang-migrate/migrate).
* Tests for the `Datastore` were implemented using [dockertest](https://github.com/ory/dockertest), which starts a container from a Postgres image, runs the migrations, and then run the test.  


db/datastore.go
```diff
@@ -0,0 +1,786 @@
...
+// NewDatastore creates a new instance of a Datastore object that connects to a Postgres database
+// parsing the connString.
+// The connString can be a URL, e.g, "postgresql://host...", or a DSN, e.g., "host= user= password= dbname= port="
+func NewDatastore(logger logrus.FieldLogger, connString string) (*Datastore, error) {
+	c, err := pgx.ParseConfig(connString)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse Postgres Connection URL: %w", err)
+	}
+
+	db := stdlib.OpenDB(*c)
```
jufantozzi, line 34: Just to confirm: this does not use a connection pool, does it?

jufantozzi (commented):
Great stuff! Did a first pass through and left a couple of comments.

On a more high-level discussion, do you still see we having one interface to abstract the datastore layer, in which each of the different implementations (postgres, mysql, etc) will have to satisfy? I couldn't find that interface anymore.

db/datastore.go
```diff
@@ -0,0 +1,786 @@
...
+		}
+	}
+
+	response, err := bundle.ToEntity()
+	if err != nil {
+		return nil, fmt.Errorf("failed converting member model to entity: %w", err)
+	}
+
+	return response, nil
+}
```
Victorblsilveira, line 339: This suggestion involves some personal changes. Feel free to ignore it, but let me know what you think.
An important change here is that I internalize the `pgMemberID` into only the `createBundle` function since is only being used there and in your solution can exist a case where is not being used and hence you are converting. Another point is that similar solution can be applied to other cases in this same PR.
```suggestion
func (d *Datastore) CreateOrUpdateBundle(ctx context.Context, req *entity.Bundle) (*entity.Bundle, error) {
	if !req.MemberID.Valid {
		return nil, errors.New("bundle member ID cannot be empty")
	}

	bundle, err := func()(*entity.Bundle, error) {
		if req.ID.Valid {
			return d.updateBundle(ctx, req)
		} else {
			return d.createBundle(ctx, req)
		}
	}()
	
	if err != nil {
	   return nil, err
	}

	response, err := bundle.ToEntity()
	if err != nil {
		return nil, fmt.Errorf("failed converting member model to entity: %w", err)
	}

	return response, nil
}

func (d *Datastore) createBundle(ctx context.Context, req *entity.Bundle) (*entity.Bundle, error) {
	pgMemberID, err := uuidToPgType(req.MemberID.UUID)
	if err != nil {
		return nil, err
	}

	params := CreateBundleParams{
		RawBundle:    req.RawBundle,
		Digest:       req.Digest,
		SignedBundle: req.SignedBundle,
		TlogID:       req.TlogID,
		SvidPem: sql.NullString{
			String: req.SvidPem,
			Valid:  true,
		},
		MemberID: pgMemberID,
	}

	bundle, err = d.querier.CreateBundle(ctx, params)
	if err != nil {
		return nil, fmt.Errorf("failed creating new bundle: %w", err)
	}

	return bundle, err
}

func (d *Datastore) updateBundle(ctx context.Context, req *entity.Bundle) (*entity.Bundle, error) {
	pgID, err := uuidToPgType(req.ID.UUID)
	if err != nil {
		return nil, err
	}
	params := UpdateBundleParams{
		ID:           pgID,
		RawBundle:    req.RawBundle,
		Digest:       req.Digest,
		SignedBundle: req.SignedBundle,
		TlogID:       req.TlogID,
		SvidPem: sql.NullString{
			String: req.SvidPem,
			Valid:  true,
		},
	}

	bundle, err = d.querier.UpdateBundle(ctx, params)
	if err != nil {
		return nil, fmt.Errorf("failed updating bundle: %w", err)
	}

	return bundle, err
}
```

db/datastore_test.go
```diff
@@ -0,0 +1,1032 @@
...
+	dbname        = "test_db"
+)
+
+var (
+	td1     = spiffeid.RequireTrustDomainFromString("foo.test")
+	td2     = spiffeid.RequireTrustDomainFromString("bar.test")
+	otherTd = spiffeid.RequireTrustDomainFromString("other.test")
+)
+
+func TestCRUDFederationGroup(t *testing.T) {
```
Victorblsilveira, line 36: Does worth it to split the tests cases in here?

Victorblsilveira (commented):
Amazing work!!! See if my comments make sense

Victorblsilveira:
> Great stuff! Did a first quick pass-through and got stuck thinking into two different things:
> 
> On a more high-level discussion, do you still see we having one interface to abstract the datastore layer, in which each of the different implementations (postgres, mysql, etc) will have to satisfy? I couldn't find that interface anymore, but the `Datastore` object looks like it would implement this interface.
> 
> and the 2nd topic is related to using a connection pool, can be discussed in the comment below.

About the abstract datastore layer, I think that is this one: [querier.go](https://github.com/HewlettPackard/galadriel/blob/fddd152e317e84689410c7e4d203c886cb16a370/db/querier.go)

db/datastore.go
jufantozzi: I'm in favor of refactoring some of these into smaller functions to improve readability, specially the larger ones.

db/datastore.go
```diff
@@ -0,0 +1,786 @@
...
+	case err != nil:
+		return nil, fmt.Errorf("failed looking up bundle with ID=%q: %w", bundleID, err)
+	}
+
+	r, err := m.ToEntity()
+	if err != nil {
+		return nil, fmt.Errorf("failed converting model bundle to entity: %w", err)
+	}
+
+	return r, nil
```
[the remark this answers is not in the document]
maxlambrecht: Yes, I was planning to refactor that into smaller functions.

db/schema.go
```diff
@@ -0,0 +1,46 @@
...
+	"database/sql"
+	"embed"
+
+	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/database/postgres"
+	"github.com/golang-migrate/migrate/v4/source/iofs"
+)
+
+//go:embed migrations/*.sql
+var fs embed.FS
```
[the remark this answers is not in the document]
Victorblsilveira: Embed go feature allows embedding files in compile time. However, some errors can only be caught in runtime. (ie. If a SQL file is not well formatted or even when there is no SQL in an empty file). Another similar feature is `go:generate` which allows running code at compile time. Take a look: 
- https://blog.carlmjohnson.net/post/2016-11-27-how-to-use-go-generate/
- https://blog.carlmjohnson.net/post/2021/how-to-use-go-embed/

db/migrations/1_initialize_schema.up.sql
```diff
@@ -0,0 +1,85 @@
...
+    digest        BYTEA                    NOT NULL,
+    signed_bundle BYTEA,
+    tlog_id       UUID,
+    svid_pem      TEXT,
+    member_id     UUID                     NOT NULL UNIQUE,
+    created_at    TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+    updated_at    TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now()
+);
+
+CREATE TABLE join_tokens
```
jufantozzi, line 48: Will join tokens not be related to harvesters in any way? Thinking about HA, does it matter if we're able to validate if the join token was for a specific harvester?

jufantozzi: I know this has some impacts we would also have to think about, like how to deal when a harvester is deleted. This specifically is also something I was thinking for a bit and considering: wouldn't it be a better idea to instead of deleting harvesters to just disable them? is this data really not worth keeping?

db/schema.go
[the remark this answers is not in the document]
maxlambrecht: The embedded SQLs (for migrations) are being exercised in the CRUD tests.

db/migrations/1_initialize_schema.up.sql
Victorblsilveira: According to our data model, there is an implicit relation between join_token and harvesters through members relation, but not an explicit one. Would be enough for the validation to know that the join token is for a member that is related to the harvester, or you are thinking of a more restricted validation in which the token needs to be exactly issued for a harvester?

maxlambrecht: > Would be enough for the validation to know that the join token is for a member that is related to the harvester?

As I understand it so far, I think it would be enough for the validation knowing which member a token belongs, whether it was used or not, and whether it didn't expire.

jufantozzi: > According to our data model, there is an implicit relation between join_token and harvesters through members relation, but not an explicit one

my bad on the wording, that's what I actually meant 😅

db/datastore_test.go
maxlambrecht: I like this tests pattern for the CRUD operations as it avoids a lot of duplication. It's for testing the happy path of an entity from creation through deletion. Then if we need to test specific scenarios or error cases I find it worthy adding specific test methods.

db/datastore.go
maxlambrecht: Sure, I was planning to do it, only it was not a priority before submitting a draft PR.

maxlambrecht:
> > Great stuff! Did a first quick pass-through and got stuck thinking into two different things:
> > On a more high-level discussion, do you still see we having one interface to abstract the datastore layer, in which each of the different implementations (postgres, mysql, etc) will have to satisfy? I couldn't find that interface anymore, but the `Datastore` object looks like it would implement this interface.
> > and the 2nd topic is related to using a connection pool, can be discussed in the comment below.
> 
> About the abstract datastore layer, I think that is this one: [querier.go](https://github.com/HewlettPackard/galadriel/blob/fddd152e317e84689410c7e4d203c886cb16a370/db/querier.go)

We could still abstract the Datastore layer in an interface that uses entity objects from the API. I didn't think about that because this PR introduces a Datastore that just supports one dialect, and I still don't know if we'll need that abstraction anyway, even in the case of supporting multiple dialects. 
The `querier` interface is auto-generated by the `sqlc` and it's a lower level interface that uses DB models.

jufantozzi:
> We could still abstract the Datastore layer in an interface that uses entity objects from the API. I didn't think about that because this PR introduces a Datastore that just supports one dialect, and I still don't know if we'll need that abstraction anyway, even in the case of supporting multiple dialects.
The querier interface is auto-generated by the sqlc and it's a lower level interface that uses DB models.

Right, it makes sense. Yes, the querier interface looks good enough, at least for now. Although this higher level abstraction will still be implemented in some form, (at the handlers, most likely?) by having an interface we're better segregating this logic for our handlers. Anyways, not something for right now

db/datastore.go
maxlambrecht: I did some refactors to reduce the complexity of the the `CreateOrUpdate` methods.

Victorblsilveira: They look great.

jufantozzi (commented):
LGTM!! Btw, you're not planning on incorporating these changes into galadriel on this PR, right?  If not, I'm ready to approve.

maxlambrecht:
> LGTM!! Btw, you're not planning on incorporating these changes into galadriel on this PR, right? If not, I'm ready to approve.

No need to approve just yet, I'm expecting more feedback from the rest of the team.

jufantozzi:
>  Btw, you're not planning on incorporating these changes into galadriel on this PR, right?

Just realized my last question was a bit ambiguous, I wanted to ask about replacing the current In memory datastore to use this implementation. This should happen in a separate PR right?

maxlambrecht:
> > Btw, you're not planning on incorporating these changes into galadriel on this PR, right?
> 
> Just realized my last question was a bit ambiguous, I wanted to ask about replacing the current In memory datastore to use this implementation. This should happen in a separate PR right?

IMO migrating Galadriel to the new Datastore is not worth doing right now, since the whole implementation of Galadriel will be mostly re-visited while moving from the POC to the MVP. Each new (re)implemented functionality in Galadriel Server will need to use the new Datastore.

db/migrations/1_initialize_schema.up.sql
```diff
@@ -0,0 +1,85 @@
...
+    status      status                   NOT NULL,
+    created_at  TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+    updated_at  TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now()
+);
+
+CREATE TABLE members
+(
+    id           UUID PRIMARY KEY                  DEFAULT gen_random_uuid(),
+    trust_domain TEXT                     NOT NULL UNIQUE,
+    status       status                   NOT NULL,
```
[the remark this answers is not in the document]
maxlambrecht: So, `member_status` and `membership_status`?

db/datastore.go
```diff
@@ -0,0 +1,889 @@
...
+
+	r, err := m.ToEntity()
+	if err != nil {
+		return nil, fmt.Errorf("failed converting model member to entity: %w", err)
+	}
+
+	return r, nil
+}
+
+func (d *Datastore) FindMemberByTrustDomain(ctx context.Context, trustDomain spiffeid.TrustDomain) (*entity.Member, error) {
```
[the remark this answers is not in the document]
maxlambrecht: This datastore implementation doesn't have Organizations

db/datastore.go
```diff
@@ -0,0 +1,889 @@
...
+		}
+		result[i] = r
+	}
+
+	return result, nil
+}
+
+func (d *Datastore) CreateOrUpdateBundle(ctx context.Context, req *entity.Bundle) (*entity.Bundle, error) {
+	if !req.MemberID.Valid {
+		return nil, errors.New("bundle member ID cannot be empty")
```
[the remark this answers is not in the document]
maxlambrecht: hmm the `MemberID` in the Bundle entity should be a `UUID`, not a `NullUUID`, as the `MemberID` cannot be null, so this if can be removed. 
I'll do this refactor also for other entities that have `NullUUID`.

pkg/server/endpoints/harvester.go
```diff
@@ -31,21 +47,20 @@ func (e *Endpoints) postBundleHandler(ctx echo.Context) error {
...
+	err = json.Unmarshal(body, &harvesterReq)
 	if err != nil {
 		e.handleTcpError(ctx, fmt.Sprintf("failed to unmarshal state: %v", err))
 		return err
 	}
 
-	if receivedHarvesterState.TrustDomain.Compare(token.TrustDomain) != 0 {
-		err := fmt.Errorf("authenticated trust domain {%s} does not match received trust domain {%s}", receivedHarvesterState.TrustDomain.String(), token.TrustDomain.String())
-
+	if harvesterReq.TrustDomainName != authenticatedTD.Name {
```
jufantozzi, line 57: Just out of curiosity, was this change due to readability?

pkg/server/endpoints/harvester.go
```diff
@@ -148,52 +207,66 @@ func (e *Endpoints) syncFederatedBundleHandler(ctx echo.Context) error {
...
+		b, err := e.Datastore.FindBundleByTrustDomainID(ctx, id)
+		if err != nil {
+			return nil, nil, err
+		}
+		td, err := e.Datastore.FindTrustDomainByID(ctx, id)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if b != nil {
```
jufantozzi, line 263: Just wondering if we might want to have a debug(?) log in case this is nil

jufantozzi (commented):
Amazing work @maxlambrecht ❤️ 

Did a code pass-through and it looks impeccable, I'm just not approving it now cause I didn't have time to test yet. 

One topic that came to my mind though, is that I noticed we're not supporting TLS for the postgres connection. IDK if this is something needed for now, I don't think we've discussed this yet, but I'm pretty sure we'll have to think about this at some point. WDYT?

pkg/server/endpoints/harvester.go
maxlambrecht: Yes, IMO it reads better that way.

maxlambrecht:
> Amazing work @maxlambrecht ❤️
> 
> Did a code pass-through and it looks impeccable, I'm just not approving it now cause I didn't have time to test yet.
> 
> One topic that came to my mind though, is that I noticed we're not supporting TLS for the postgres connection. IDK if this is something needed for now, I don't think we've discussed this yet, but I'm pretty sure we'll have to think about this at some point. WDYT?

Thanks Juliano! ❤️ 

The Postgres connection string allows TLS, something like: 

```
"sslmode=require sslrootcert=server-ca.pem sslcert=client-cert.pem sslkey=client-key.pem hostaddr=your_host port=5432 user=your_user dbname=your_db"
```

jufantozzi (approved):
💯 💯

mgbcaio (commented):
this is a number 10 work, right here. Nice job, Max 🏅 Ready to approve as soon as others take a look into this.

pkg/server/endpoints/harvester.go
```diff
@@ -89,13 +116,25 @@ func (e *Endpoints) postBundleHandler(ctx echo.Context) error {
 func (e *Endpoints) syncFederatedBundleHandler(ctx echo.Context) error {
 	e.Logger.Debug("Receiving sync request")
 
-	token, ok := ctx.Get("token").(*common.AccessToken)
+	t, ok := ctx.Get("token").(*entity.JoinToken)
```
Victorblsilveira, line 119: Name suggestions for variable `t`: 
- `jt`
- `joinToken`

pkg/server/endpoints/harvester.go
```diff
@@ -18,9 +20,23 @@ import (
 func (e *Endpoints) postBundleHandler(ctx echo.Context) error {
 	e.Logger.Debug("Receiving post bundle request")
 
-	token, ok := ctx.Get("token").(*common.AccessToken)
+	t, ok := ctx.Get("token").(*entity.JoinToken)
```
Victorblsilveira, line 23: Name suggestions for variable `t`:
- `jt`
- `joinToken`

pkg/server/endpoints/harvester.go
Victorblsilveira, line 119: What do you think about creating a const variable for the `token` and using it instead of a hard-coded string?

pkg/server/endpoints/management.go
```diff
@@ -140,35 +183,60 @@ func (e *Endpoints) listRelationshipsHandler(w http.ResponseWriter, r *http.Requ
 	}
 }
 
+func (e *Endpoints) populateTrustDomainNames(ctx context.Context, rels []*entity.Relationship) ([]*entity.Relationship, error) {
```
Victorblsilveira, line 186: Name suggestions for variable `rels`: 
- `relationships`

Victorblsilveira (approved):
Hey @maxlambrecht, great work here. 
A lot of changes, most of my comments are about naming. 
Don't feel obliged to accept it, but let me know what you think about it. 
Feel free to ignore them as well.

pkg/server/datastore/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+`sqlc` is configured through the file [sqlc.yalm](sqlc.yaml).
+
+When there is a change in the schema or in the queries, the DB code should be re-generated:
+
+```
+sqlc generate
+```
+
+This regenerates the files `models.go`, `db.go`, `querier.go`, and files ending in `.sql.go`.
```
wibarre, line 13: I noticed that the .sql.go files are in different folders. Any reason why sqlc would place them in different places? Do we want to keep them under a consistent folder structure?

pkg/server/datastore/datastore.go
```diff
@@ -2,319 +2,624 @@ package datastore
...
+	trustDomains, err := d.querier.ListTrustDomains(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting trust domain list: %w", err)
+	}
+
+	result := make([]*entity.TrustDomain, len(trustDomains))
+	for i, m := range trustDomains {
+		r, err := m.ToEntity()
+		if err != nil {
+			return nil, fmt.Errorf("failed converting model trust domain to entity: %v", err)
```
wibarre, line 177: Total ignorance from my part. Shouldn't these errors trigger a log entry via the standardized library we are using (logrus)?

pkg/server/datastore/datastore_test.go
```diff
@@ -1 +1,739 @@
...
+
+	// List tokens
+	tokens, err = datastore.ListJoinTokens(ctx)
+	require.NoError(t, err)
+	assert.Equal(t, 3, len(tokens))
+	require.Contains(t, tokens, token1)
+	require.Contains(t, tokens, token2)
+	require.Contains(t, tokens, token3)
+
+	// Update join token
```
wibarre, line 622: if I understood the datastore logic for updating the join tokens, only values of "true" can be use to update the used field. Thus, we cannot update the used field of a join token to false. That validation makes sense to me. Should we add a test case to  validate that?

pkg/server/datastore/migrations/1_initialize_schema.down.sql
```diff
@@ -0,0 +1,5 @@
+DROP TABLE IF EXISTS join_tokens;
+DROP TABLE IF EXISTS relation_ships;
```
wibarre, line 2: instead of relation_ships shouldn't this be relationships?

wibarre (commented):
Just focused on the pkg/server/datastore. Thank you Maxl for this. Looks great!

pkg/server/datastore/README.md
maxlambrecht: All the `.go` files are the datastore folder, the `.sql.go` files too. 
What are in sub-folders are the `.sql` files, to separate SQL files from Go files, and I also separated the queries definitions and the schema definitions.

pkg/server/datastore/datastore.go
```diff
@@ -2,319 +2,624 @@ package datastore
...
+	querier Querier
+}
 
-	GenerateAccessToken(ctx context.Context, t *common.AccessToken, trustDomain string) (*common.AccessToken, error)
-	GetAccessToken(ctx context.Context, token string) (*common.AccessToken, error)
-	ListRelationships(ctx context.Context) ([]*common.Relationship, error)
+// NewSQLDatastore creates a new instance of a Datastore object that connects to a Postgres database
+// parsing the connString.
+// The connString can be a URL, e.g, "postgresql://host...", or a DSN, e.g., "host= user= password= dbname= port="
+func NewSQLDatastore(logger logrus.FieldLogger, connString string) (*SQLDatastore, error) {
```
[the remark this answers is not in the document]
maxlambrecht: The Postgres connection string allows TLS, something like:

```
"sslmode=require sslrootcert=server-ca.pem sslcert=client-cert.pem sslkey=client-key.pem hostaddr=your_host port=5432 user=your_user dbname=your_db"
```

pkg/server/datastore/datastore.go
maxlambrecht: "Log and return the error" is an anti-pattern. It's better to return the error to the client and let it decide what to do with it.

pkg/server/datastore/datastore_test.go
maxlambrecht: The  method to update a join token only takes a boolean because "used" is the only field that makes sense that it be updated. But the datastore doesn't validate that it be "true" or "false", as it is a "business" logic that will be part of the Server controller.

demo/README.md
```diff
@@ -29,4 +29,4 @@ The `greeter` client fetches its SVID from the SPIRE Agent serving `two.org`, an
 
 ## Cleanup
 
-Run `x-cleanup.sh` to remove from the system all the SPIRE data created by this demo.
\ No newline at end of file
+Run `x-cleanup.sh` to remove from the system all the SPIRE bundle created by this demo.
```
mchurichi, line 32: ?

mchurichi (commented):
This is looking great, Max! 💯

pkg/server/datastore/datastore.go
```diff
@@ -2,319 +2,624 @@ package datastore
...
+}
 
-	GenerateAccessToken(ctx context.Context, t *common.AccessToken, trustDomain string) (*common.AccessToken, error)
-	GetAccessToken(ctx context.Context, token string) (*common.AccessToken, error)
-	ListRelationships(ctx context.Context) ([]*common.Relationship, error)
+// NewSQLDatastore creates a new instance of a Datastore object that connects to a Postgres database
+// parsing the connString.
+// The connString can be a URL, e.g, "postgresql://host...", or a DSN, e.g., "host= user= password= dbname= port=".
+func NewSQLDatastore(logger logrus.FieldLogger, connString string) (*SQLDatastore, error) {
+	c, err := pgx.ParseConfig(connString)
```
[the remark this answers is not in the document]
maxlambrecht: The server.conf has a comment that kind of implies it:

```
# db_conn_string: postgres database connection string.
```

pkg/server/datastore/datastore.go
```diff
@@ -2,319 +2,624 @@ package datastore
...
+	return nil
+}
+
+func (d *SQLDatastore) FindJoinToken(ctx context.Context, token string) (*entity.JoinToken, error) {
+	joinToken, err := d.querier.FindJoinToken(ctx, token)
+	switch {
+	case errors.Is(err, sql.ErrNoRows):
+		return nil, nil
+	case err != nil:
+		return nil, fmt.Errorf("failed looking up join token %q: %w", token, err)
```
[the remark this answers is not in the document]
maxlambrecht: it shouldn't be included, good catch

demo/README.md
maxlambrecht: ouch, that shouldn't have been committed

mchurichi (approved):
🌭

jufantozzi (approved):

PR #77, opened 2023-01-17
[ATB-157] Implement Telemetry package - boilerplate

mgbcaio:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**
Harvester run and Galadriel Server run functions were updated to also run the telemetry tasks, starting the Prometheus server.

**Description of change**
Add a new package related to telemetry with some boilerplate to be used with the fuctions we want to keep track of. Also added the Proemtheus configuration and runner to be used in the future.

https://jira-pro.its.hpecorp.net:8443/browse/ATB-157


pkg/common/telemetry/prometheus_test.go
```diff
@@ -0,0 +1,100 @@
...
+		errCh <- pr.run(context.Background())
+	}()
+
+	// It doesn't run if it's not configured
+	select {
+	case err := <-errCh:
+		assert.Nil(t, err, "should be nil if not configured")
+	case <-time.After(time.Minute):
+		t.Fatal("prometheus running but not configured")
+	}
```
Victorblsilveira, line 75: Here you have two separate tests running like one. Same example as above

Victorblsilveira (commented):
Hey, @mgbcaio great changes here. I hope that my comments can provide some useful feedback.

pkg/common/telemetry/prometheus_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	pr, err := newTestPrometheusRunner(config)
+	require.NoError(t, err)
+	assert.True(t, pr.isConfigured())
+
+	config.PrometheusConf = nil
+	pr, err = newTestPrometheusRunner(config)
+	require.NoError(t, err)
+	assert.False(t, pr.isConfigured())
+}
```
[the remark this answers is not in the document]
mgbcaio: yeah, i will split it up. thanks!

pkg/common/telemetry/prometheus.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if runner.c == nil {
+		return runner, nil
+	}
+
+	var err error
+	runner.sink, err = prommetrics.NewPrometheusSinkFrom(prommetrics.PrometheusOpts{})
+	if err != nil {
+		return runner, err
+	}
```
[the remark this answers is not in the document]
mgbcaio: Yeah true, but all I want from the sink its assigned it to the runner. Doesn’t it make sense doing like I did, so there isnt two different assignements ? Does it make harder to interpret ?

pkg/common/telemetry/metrics.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// ListenAndServe starts the metrics process
+func (m *MetricsImpl) ListenAndServe(ctx context.Context) error {
+	var err error
+	for _, runner := range m.runners {
+		err = util.RunTasks(ctx, runner.run)
+	}
+	return err
+}
```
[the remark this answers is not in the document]
mgbcaio: Hey victor, could you elaborate a bit more about what errors you think we may be overwriting here, I think you lost me hehe ? AFAIK, the `RunTasks` function will send the error of the spread of tasks it received, regarding of what hapenned, to the err channel and then return it, same if the context is `done`, so we would be getting the final message here.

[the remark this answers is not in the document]
Victorblsilveira: My point is that you are iterating over a list of runners 
```
for _, runner := range m.runners
```
and running a task for each. 

Supposing that the list has more than one element: 

If the first runner returns an error, and the last one also returns an error the variable `err`
will end up with the last error returned and you will be losing any other error that came before.
```
err = util.RunTasks(ctx, runner.run)
```
We can also make a call to talk about that.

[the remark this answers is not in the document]
mgbcaio: I see now. I left it like this to support multiple metric collectors, since we will only use Prometheus for now I think I will refactor that so it becomes simpler. sounds good ?

pkg/common/telemetry/prometheus.go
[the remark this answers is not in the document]
Victorblsilveira: I just present you with another implementation option. The difference between them is that in one you need to explicitly declare the error variable to be used and also independently of an error you are assigning to the sink. If don't make sense for your approach you can ignore it.

pkg/common/telemetry/metrics.go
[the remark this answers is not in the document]
Victorblsilveira: Sounds great.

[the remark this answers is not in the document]
maxlambrecht: `util.RunTasks` is a variadic function, so you can pass any number of tasks:

```
func (m *MetricsImpl) ListenAndServe(ctx context.Context) error {
	var tasks []func(context.Context) error
	for _, runner := range m.runners {
		tasks = append(tasks, runner.run)
	}

	return util.RunTasks(ctx, tasks...)
}
```

pkg/common/telemetry/metrics.go
```diff
@@ -0,0 +1,159 @@
...
+
+//SetGauge sets the gauge for the metrics
+func (m *MetricsImpl) SetGauge(key []string, val float32) {
+	for _, s := range m.metricsSinks {
+		s.SetGauge(key, val)
+	}
+}
+
+// SetGaugeWithLabels sets the gauge with labels/tags
+// TODO: Sanatize Labels
```
maxlambrecht, line 103: Label could be sanitized at creation, so a `Label` instance is always a "correct" label.

pkg/common/telemetry/prometheus.go
```diff
@@ -0,0 +1,94 @@
...
+	sink   Sink
+}
+
+func newPrometheusRunner(c *MetricsConfig) (sinkRunner, error) {
+	runner := &prometheusRunner{
+		c:   c.PrometheusConf,
+		log: c.Logger,
+	}
+
+	if runner.c == nil {
```
maxlambrecht, line 30: Isn't this an error scenario? Why return the `runner` if it has a nil config?

pkg/common/telemetry/prometheus.go
```diff
@@ -0,0 +1,94 @@
...
+		return runner, nil
+	}
+
+	var err error
+	runner.sink, err = prommetrics.NewPrometheusSinkFrom(prommetrics.PrometheusOpts{})
+	if err != nil {
+		return runner, err
```
maxlambrecht, line 37: Why return the runner in case of error?

pkg/common/telemetry/prometheus.go
mgbcaio: To keep access to the logger it has. But i do not have a strong opinion about it. just figured it might be important later on.

pkg/common/telemetry/prometheus.go
mgbcaio: same as above.

pkg/common/telemetry/config.go
```diff
@@ -0,0 +1,26 @@
+package telemetry
+
+import "github.com/sirupsen/logrus"
+
+// MetricsConfig represents metrics configuration
+type MetricsConfig struct {
+	Logger logrus.FieldLogger
+
+	// ServiceName is the name of the service that is been monitored
```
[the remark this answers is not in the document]
mgbcaio: Nice catch!

Victorblsilveira (approved):
LGTM 🥇

pkg/common/telemetry/call.go
```diff
@@ -0,0 +1,53 @@
...
+// AddLabel adds a label to be emitted with the call counter.
+func (c *CallCounter) AddLabel(name, value string) {
+	c.mu.Lock()
+	c.labels = append(c.labels, Label{Name: name, Value: value})
+	c.mu.Unlock()
+}
+
+// Done finishes the "call" and emits metrics.
+func (c *CallCounter) Done(errp *error) {
+	if c.done {
```
maxlambrecht, line 39: Why don't you use the Mutex in this method? I seems to me that there could be a race condition with the `c.done`.

pkg/common/telemetry/prometheus.go
maxlambrecht: I think we need to `return nil, someError`

pkg/common/telemetry/prometheus.go
maxlambrecht: Same as above :)

pkg/common/telemetry/prometheus.go
Victorblsilveira: Couldn't you just check the c.PrometheusConf before assigning it to the structure and returning an error due to the requirement of the configuration? You still have access to the log in function which will call `newPrometheusRunner`.
So there is no strong reason for returning the sink as @maxlambrecht  point out.

pkg/common/telemetry/call.go
mgbcaio: I didnt thoght of a use case where the `Done` function gets called concurrently to actually have the `done`field unsynchronized, thats why I didnt use the mutex lock. But I think I will use it to avoid any problems in the future.

mgbcaio:
Closing this for now.

PR #94, opened 2023-01-23
[FIX] Handling XSS and log injection 

Victorblsilveira:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

Fixing some alerts raised by our security analysis bot.


mgbcaio (approved):
🥇

pkg/server/endpoints/harvester.go
```diff
@@ -79,7 +79,9 @@ func (e *Endpoints) postBundleHandler(ctx echo.Context) error {
 			e.handleTcpError(ctx, fmt.Sprintf("failed to update member: %v", err))
 			return err
 		}
-		e.Logger.Debugf("Trust domain %s has been successfully updated", receivedHarvesterState.TrustDomain)
+
+		td := util.LogSanitize(receivedHarvesterState.TrustDomain.String())
```
maxlambrecht, line 83: I have two questions:
- Why do we need to Sanitize the String representation of a parsed TrustDomain?
- Is there a way to enable log sanitization at the "library level" instead of doing it on each log line?

Victorblsilveira: > I have two questions:
> 
> * Why do we need to Sanitize the String representation of a parsed TrustDomain?
> * Is there a way to enable log sanitization at the "library level" instead of doing it on each log line?

1- The Trust Domain string is directly coming from the request without any proper sanitizing. This could turn into a log injection issue if not handled properly, even if it's parsed.
2- This is a great idea, but as far as I look it into our log library at the time I was fixing I didn't see anything to help with that.

maxlambrecht: The TrustDomain string that comes in the request is parsed to a `spiffe.TrustDomain` object, see `TrustBundle`

```
type TrustBundle struct {
	// Trust Domain of the bundle
	TrustDomain spiffeid.TrustDomain `json:"trust_domain"`
...
```

maxlambrecht: > 
> 1- The Trust Domain string is directly coming from the request without any proper sanitizing. This could turn into a log injection issue if not handled properly, even if it's parsed. 


The TrustDomain string that comes in the request is parsed to a `spiffe.TrustDomain` object, see `TrustBundle`

```
type TrustBundle struct {
	// Trust Domain of the bundle
	TrustDomain spiffeid.TrustDomain `json:"trust_domain"`
...
```

Victorblsilveira: I know it, but there is no string validation, could be anything.

Victorblsilveira: > It cannot be anything, the `TrustDomain.parse()` applies all these restrictions: https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md#21-trust-domain
> 
> Are those restrictions not sufficient?

It seems to be. But GitHub alerts still notify of that issue. What would we do in that case? just ignore it?

maxlambrecht: It's a false positive, what else should we do other than conscientiously ignore it? Why would we be add code just to make the CodeQL happy?

Victorblsilveira: The intent here is not to make CodeQL happy but potentially fix known problems if it's a false positive, I think we can ignore it, but is there a way to signalize that false positive to the CodeQL community for potential fixes? Or even there is a way to overcome spiffe parse and inject log that we don't know why? For now, I think I can revert the change.
Quick question, does it worth it to keep any of the trust domain sanitizing?

Victorblsilveira: > The intent here is not to make CodeQL happy but potentially fix known problems if it's a false positive, I think we can ignore it, but is there a way to signalize that false positive to the CodeQL community for potential fixes? Or even there is a way to overcome spiffe parse and inject log that we don't know why? For now, I think I can revert the change. Quick question, does it worth it to keep any of the trust domain sanitizing?

I find a way to mark it as a false positive in the GitHub alert. 🚀

maxlambrecht: I understand your point, I know the intent is not to make CodeQL happy, but in this particular case, in which we know we are not logging raw input strings directly but we are using a type with strong restrictions cooked-in, adding code that CodeQL is able to detect as sanitisation seems to me like we are trying to make CodeQL happy. 

There's already a similar issue reported in CodeQL repo: https://github.com/github/codeql/issues/10922

maxlambrecht: > > The intent here is not to make CodeQL happy but potentially fix known problems if it's a false positive, I think we can ignore it, but is there a way to signalize that false positive to the CodeQL community for potential fixes? Or even there is a way to overcome spiffe parse and inject log that we don't know why? For now, I think I can revert the change. Quick question, does it worth it to keep any of the trust domain sanitizing?
> 
> I find a way to mark it as a false positive in the GitHub alert. 🚀 See: [Log entries created from user input](/HewlettPackard/galadriel/security/code-scanning/30)

Awesome!! 🚀

mgbcaio (approved):

maxlambrecht (approved):
💯

PR #107, opened 2023-02-27
implements etcd connection

jufantozzi:
**Pull request check list**

- [ ] Documentation updated?

**Description of change**
Adds initial code for an etcd database connection hook in the Harvester. The configuration structure used by this package was stripped and placed into the same file in an attempt to maintain low impact in our current code, as this is implementation will not be happening in a short future.

Additionally, a draft HA support code for bundle pushing was implemented. All this logic is commented out in the `bundle.go` file. 
I'm open to remove this code from the PR if team thinks it may add too much noise to current code, but should be good to at least have some other eyes on this already.
 
This PR does not alter the current harvester behavior in any way.


jufantozzi:
current discussions are leading to a no-database architecture for the Harvester, closing this for now

PR #116, opened 2023-03-10
Initial GCA implementation

maxlambrecht:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**

Initial Galadriel CA boilerplate and Signing functionality for JWT tokens and X.509 certificates.

**Description of change**

Adds the Galadriel CA component, with two endpoints: one on listening on UDS, the other listening on TCP using TLS, with a certificate generated and kept in memory, and rotated every hour. 

The GCA needs to be configured with a `root_cert.pem` and `root_key.pem`. This Root CA is used for signing the X.509 and JWT tokens.

The GCA uses the new interface `ca.ServerCA` to sign certificates and JWT tokens. This same interface will be used by Galadriel Server to implement the same mechanisms for TLS and for providing the Haversters with new JWT tokens. 

Adds some functionality to load, parse, and create X.509 certificates and RSA keys. 


pkg/gca/endpoints/jwt/handler.go
```diff
@@ -0,0 +1,119 @@
...
+	CA          ca.ServerCA
+	Logger      logrus.FieldLogger
+	JWTTokenTTL time.Duration
+	Clock       clock.Clock
+}
+
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	h.Logger.Debug("New JWT authToken requested")
+
+	authHeader := r.Header.Get("Authorization")
```
Victorblsilveira, line 31: Can we create a constant for "Authorization"?

pkg/gca/endpoints/jwt/handler.go
```diff
@@ -0,0 +1,119 @@
...
+	if authHeader == "" {
+		http.Error(w, "Authorization header is missing", http.StatusBadRequest)
+		return
+	}
+	if !strings.HasPrefix(authHeader, "Bearer ") {
```
Victorblsilveira, line 36: "Bearer " constant also?

pkg/gca/endpoints/jwt/handler.go
```diff
@@ -0,0 +1,119 @@
...
+		http.Error(w, "Invalid Authorization header format", http.StatusBadRequest)
+		return
+	}
+
+	// Extract the JWT from the header
+	bearerToken := strings.TrimPrefix(authHeader, "Bearer ")
```
Victorblsilveira, line 42: "Bearer " constant?

pkg/gca/endpoints/jwt/handler.go
```diff
@@ -0,0 +1,119 @@
...
+	found := false
+	for _, a := range aud {
+		if a == jwtAudience {
+			found = true
+			break
+		}
+	}
+
+	return found
+}
```
Victorblsilveira, line 119: Simplify.
```suggestion
func containsAudience(aud []interface{}) bool {
	for _, a := range aud {
		if a == jwtAudience {
			return true
		}
	}
	return false
}
```

Victorblsilveira (commented):
Hey, @maxlambrecht a lot of work here. 
Great job. 
Most of my comments are just observations and simplifications.

The biggest concern that I have here is that you are committing certificates, even if they are for testing purposes, they tend to expire and cause the unit test to fail. I know that this can take time depending on the certificate's valid time, but I would say to generate new ones in each unit test run or document the process for generating these certificates since they can use specific configurations or not, depending on the case.

Let me know WDYT.

pkg/gca/endpoints/endpoints.go
```diff
@@ -0,0 +1,247 @@
...
+			cert, err := e.getTLSCertificate(ctx)
+			if err != nil {
+				errChan <- fmt.Errorf("failed to rotate GCA certificate: %w", err)
+			}
+			certStore.cert = cert
+		case <-ctx.Done():
+			e.Logger.Info("Stopped GCA Certificate rotator")
+			return
+		}
+	}()
```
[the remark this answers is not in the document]
maxlambrecht: Done

maxlambrecht:
> 
> The biggest concern that I have here is that you are committing certificates, even if they are for testing purposes, they tend to expire and cause the unit test to fail. I know that this can take time depending on the certificate's valid time, but I would say to generate new ones in each unit test run or document the process for generating these certificates since they can use specific configurations or not, depending on the case.
> 
> Let me know WDYT.

Yeah, I agree. I removed the certificate files for the tests of the CA and Endpoints, they are now generated ad-hoc in memory. The only ones that are left are used for testing the "load certs and keys from disk" methods.

pkg/common/ca/ca_test.go
```diff
@@ -0,0 +1,131 @@
...
+	assert.False(t, cert.IsCA)
+	assert.True(t, cert.BasicConstraintsValid)
+	assert.Equal(t, config.Clock.Now().Add(-NotBeforeTolerance), cert.NotBefore)
+	assert.Equal(t, config.Clock.Now().Add(oneMinute), cert.NotAfter)
+	assert.Equal(t, cert.KeyUsage, x509.KeyUsageKeyEncipherment|
+		x509.KeyUsageKeyAgreement|
+		x509.KeyUsageDigitalSignature)
+	assert.Equal(t, cert.ExtKeyUsage, []x509.ExtKeyUsage{
+		x509.ExtKeyUsageServerAuth,
+		x509.ExtKeyUsageClientAuth})
```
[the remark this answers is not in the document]
maxlambrecht: I tried, but a certificate object you create with `&x509.Certificate{}` will have several fields with `nil`, e.g. `Raw`, `Signature`, etc, that makes impossible to compare the two certificate objects.

pkg/gca/endpoints/endpoints.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	mux.Handle("/jwt", jwtHandler)
+
+	server := &http.Server{
+		Addr:      e.TCPAddress.String(),
+		Handler:   mux,
+		TLSConfig: tlsConfig,
+	}
+
+	e.Logger.Infof("Starting GCA TCP endpoint listening on %s", e.TCPAddress.String())
```
jufantozzi, line 160:
```suggestion
	e.Logger.Infof("Starting secure GCA TCP endpoint listening on %s", e.TCPAddress.String())
```

jufantozzi (approved):
This approve will most likely get staled, but goes to show this LGTM already! Amazing work @maxlambrecht!

I'll leave one question, though:

I've noticed you're using `go-jose` instead of `go-jwt`. Doing some research, seems like `go-jwt` is way more popular, but it's older and is decreasing in activity, while `go-jose` looks like it has been actively worked on. Is this, at least, partially, the reasoning to chose from either one? Would like to hear your thoughts on this

cmd/gca/cli/config_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	assert.Equal(t, "127.0.0.1", sc.TCPAddress.IP.String())
+	assert.Equal(t, config.GCA.ListenPort, sc.TCPAddress.Port)
+	assert.Equal(t, config.GCA.SocketPath, sc.LocalAddress.String())
+	assert.Equal(t, strings.ToLower(config.GCA.LogLevel), logrus.GetLevel().String())
+	assert.Equal(t, config.GCA.RootCertPath, sc.RootCertPath)
+	assert.Equal(t, config.GCA.RootKeyPath, sc.RootKeyPath)
+	assert.Equal(t, time.Duration(18000000000000), sc.X509CertTTL)
+	assert.Equal(t, time.Duration(36000000000000), sc.JWTCertTTL)
```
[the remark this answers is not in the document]
maxlambrecht: yep, what was I thinking? xD 😄

pkg/gca/gca.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// GCA is a struct that represents a Galadriel CA.
+type GCA struct {
+	config *Config
+	CA     *ca.CA
+}
+
+// NewGCA creates a new Galadriel CA GCA with the given configuration.
+func NewGCA(config *Config) (*GCA, error) {
```
jufantozzi, line 52:
```suggestion
func New(config *Config) (*GCA, error) {
```

pkg/gca/endpoints/jwt/handler.go
```diff
@@ -0,0 +1,164 @@
...
+
+		if _, err := w.Write([]byte(newToken)); err != nil {
+			handler.Logger.Errorf("error writing token in HTTP response: %w", err)
+		}
+	}), nil
+}
+
+func getAuthJWTToken(w http.ResponseWriter, r *http.Request) (*jwt.JSONWebToken, bool) {
+	authHeader := r.Header.Get(AuthorizationHeader)
+	if authHeader == "" {
```
Victorblsilveira, line 103: Do we want to trim space headers?

pkg/gca/endpoints/jwt/handler.go
```diff
@@ -0,0 +1,164 @@
...
+	return jwtToken, true
+}
+
+func (h *Handler) validateClaims(w http.ResponseWriter, claims map[string]any) bool {
+	if h.isTokenExpired(claims) {
+		http.Error(w, "expired JWT token", http.StatusUnauthorized)
+		return false
+	}
+
+	aud, ok := claims["aud"].([]any)
```
Victorblsilveira, line 129: Const for "aud", "exp", "sup" and other ones?

pkg/gca/endpoints/jwt/handler.go
```diff
@@ -0,0 +1,164 @@
...
+		http.Error(w, "invalid JWT token audience", http.StatusUnauthorized)
+		return false
+	}
+
+	return true
+}
+
+func (h *Handler) isTokenExpired(claims map[string]any) bool {
+	var expiration time.Time
+	switch exp := claims["exp"].(type) {
```
Victorblsilveira, line 145: what about casting errors here?

Victorblsilveira (commented):
Overall it's pretty good, just left some minor comments here. See if they make sense

maxlambrecht:
> I'll leave one question, though:
> 
> I've noticed you're using `go-jose` instead of `go-jwt`. Doing some research, seems like `go-jwt` is way more popular, but it's older and is decreasing in activity, while `go-jose` looks like it has been actively worked on. Is this, at least, partially, the reasoning to chose from either one? Would like to hear your thoughts on this

I chose `go-jose` because it provides the generation of `JWKs`, but after putting a little more thought into it, I don't think we're going to need that capability, because GCA is not going to distribute JWT Public Key Sets like SPIRE does. Having said that, I think it's better to use `go-jwt`, its API is higher level than `go-jose`, and, as you said, it's more popular. 

I'm migrating to`go-jwt`.

pkg/gca/endpoints/jwt/handler.go
maxlambrecht: Removed that function. I migrated to `go-jwt`, by default it validates the expiration when parsing the token.

pkg/gca/endpoints/jwt/handler.go
maxlambrecht: No need that anymore.

pkg/gca/endpoints/jwt/handler.go
```diff
@@ -0,0 +1,164 @@
...
+		if err != nil {
+			handler.Logger.WithError(err).Error("Failed to generate JWT token")
+			http.Error(w, "error generating new token", http.StatusInternalServerError)
+			return
+		}
+
+		if _, err := w.Write([]byte(newToken)); err != nil {
+			handler.Logger.Errorf("error writing token in HTTP response: %w", err)
+		}
+	}), nil
```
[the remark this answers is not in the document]
maxlambrecht: Done

pkg/gca/endpoints/endpoints.go
```diff
@@ -0,0 +1,279 @@
...
+	tcpAddress *net.TCPAddr
+	localAddr  net.Addr
+	logger     logrus.FieldLogger
+	clock      clock.Clock
+	config     *Config
+
+	certsStore *tlsCertSource
+
+	hooks struct {
+		// test hook used to indicate that is listening on TCP
```
jufantozzi, line 73: I haven't looked deep enough into this, but is this *strictly* necessary? Correct me if I'm wrong, but it looks like this is just used for testing (in `endpoints_test.go`), and by the way this is being used, feels like we can have some workaround to it. For instance, why not use something as simple as a `sleep` to wait for the server going up?

maxlambrecht:
I removed the JWT API handler from this PR, since I'll add it as part of a more specific issue. 
I left the common methods and interfaces for working with JWT tokens.

jufantozzi (commented):
Ready to approve this, but thought it was important to bring this up and want to hear your thoughts

pkg/gca/endpoints/endpoints.go
jufantozzi: Valid points, thanks for the clarification 💯

jufantozzi (approved):

pkg/common/ca/ca_test.go
```diff
@@ -0,0 +1,171 @@
...
+		TTL:      time.Minute,
+	}
+
+	token, err := serverCA.SignJWT(params)
+	require.NoError(t, err)
+	require.NotNil(t, token)
+
+	claims := &jwt.RegisteredClaims{}
+	parsed, err := jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) { return serverCA.PublicKey(), nil })
+
```
Victorblsilveira, line 97: Make it easier to read it.
```suggestion
  getServerPublicKey := func(token *jwt.Token) (interface{}, error) { return serverCA.PublicKey(), nil }
	parsed, err := jwt.ParseWithClaims(token, claims, getServerPublicKey)
	
```

pkg/common/ca/ca_test.go
```diff
@@ -0,0 +1,171 @@
...
+		TTL:          1 * time.Minute,
+		CustomClaims: customClaims,
+	}
+
+	token, err := serverCA.SignJWT(params)
+	require.NoError(t, err)
+	require.NotNil(t, token)
+
+	claims := &ca.Claims{}
+	parsed, err := jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) { return serverCA.PublicKey(), nil })
```
Victorblsilveira, line 132: Same
```suggestion
  getServerPublicKey := func(token *jwt.Token) (interface{}, error) { return serverCA.PublicKey(), nil }
	parsed, err := jwt.ParseWithClaims(token, claims, getServerPublicKey)
```

pkg/gca/endpoints/endpoints.go
```diff
@@ -0,0 +1,266 @@
...
+			e.logger.WithError(err).Error("Error closing GCA UDS endpoint")
+		}
+		<-errChan
+		e.logger.Info("GCA UDS endpoint stopped")
+
+		return nil
+	}
+}
+
+func (e *Endpoints) tlsCertificateRotator(ctx context.Context, errChan chan error) {
```
Victorblsilveira, line 211: Name change suggestion: `startTLSCertificateRotator`

pkg/gca/endpoints/endpoints.go
```diff
@@ -0,0 +1,266 @@
...
+	case <-ctx.Done():
+		e.logger.Info("Stopping GCA TCP endpoint")
+		err = server.Close()
+		if err != nil {
+			e.logger.WithError(err).Error("Error closing GCA TCP endpoint")
+		}
+		<-errChan
+		e.logger.Info("GCA TCP endpoint stopped")
+		return nil
+	}
```
Victorblsilveira, line 170: Should we close the error channel? In the case of a yes. The UDS also needs that.
```suggestion
	select {
	case err = <-errChan:
		e.logger.WithError(err).Error("GCA TCP endpoint stopped prematurely")
		close(errChan)
		return err
	case <-ctx.Done():
		e.logger.Info("Stopping GCA TCP endpoint")
		err = server.Close()
		if err != nil {
			e.logger.WithError(err).Error("Error closing GCA TCP endpoint")
		}
		<-errChan
		e.logger.Info("GCA TCP endpoint stopped")
		return nil
	}
```

pkg/gca/endpoints/endpoints.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+func (t *tlsCertSource) getCert() *tls.Certificate {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	return t.cert
+}
+
+func (e *Endpoints) runTCPServer(ctx context.Context) error {
+	var err error
```
Victorblsilveira, line 123: I think you can remove this. `err` can be defined in the lines below regarding this declaration.

Victorblsilveira (commented):
Just making another review round @maxlambrecht. See if my comments make sense.

pkg/gca/endpoints/endpoints.go
maxlambrecht: I don't think we need this, as channels in Go are not resources like files or sockets that need to be explicitly closed to free up system resources. When a channel is no longer referenced by any goroutine, it becomes eligible for garbage collection by the Go runtime, and the memory used by the channel is automatically reclaimed.

pkg/common/cryptoutil/certs_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+	"testing"
+
+	"github.com/HewlettPackard/galadriel/pkg/common/cryptoutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParsePrivateKeyPEM(t *testing.T) {
+	// not a private key
+	_, err := cryptoutil.ParseRSAPrivateKeyPEM(readFile(t, "testdata/cert.pem"))
```
mgbcaio, line 14: Can you extract these `testdata/files` to a `const` declaration? Since you repeat them throughout this file.

pkg/gca/endpoints/endpoints_test.go
```diff
@@ -0,0 +1,99 @@
...
+	config, _ := newEndpointTestConfig(t)
+
+	endpoint, err := New(config)
+	require.NoError(t, err)
+	require.NotNil(t, endpoint)
+
+	assert.Equal(t, endpoint.serverCA, config.ServerCA)
+	assert.Equal(t, endpoint.tcpAddress, config.TCPAddress)
+	assert.Equal(t, endpoint.localAddr, config.LocalAddress)
+	assert.Equal(t, endpoint.logger, config.Logger)
```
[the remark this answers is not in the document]
maxlambrecht: Yes!

Victorblsilveira (approved):
LGTM!

maxlambrecht:
Closing as this is no longer a priority.

PR #118, opened 2023-03-14
Adds OpenAPI definitions for GCA Harvester API and GCA Admin API

mchurichi:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**
N/A

**Description of change**
Adds OpenAPI definitions for GCA Harvester API and GCA Admin API.
WIP: codegen stuff, documentation

N/A

### DIFF
--- a/pkg/ca/api/admin.yaml
+++ b/pkg/ca/api/admin.yaml
@@ -1,0 +1,262 @@
+openapi: 3.0.3
+info:
+  title: Galadriel Server Admin API - OpenAPI 3.0
+  version: 1.0.11
+  description: Galadriel Server Management API specification.
+  contact:
+    name: Github
+    url: https://github.com/HewlettPackard/galadriel
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+servers:
+  - url: https://localhost.swagger.io/api/v3
+paths:
+  /trust-domains:
+    get:
+      operationId: ListTrustDomains
+      summary: Retrieves all Trust Domains
+      tags:
+        - Trust Domain
+      responses:
+        '200':
+          description: Successfull operation. Returns all Trust Domains
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/TrustDomain'
+        default:
+          $ref: '#/components/responses/Default'
+    post:
+      tags:
+        - Trust Domain
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TrustDomainRequest'
+        required: true
+      responses:
+        '201':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TrustDomain'
+          description: Trust Domain successfully created. Returns newly created Trust Domain
+        default:
+          $ref: '#/components/responses/Default'
+      operationId: CreateTrustDomain
+      summary: Creates a new Trust Domain
+  /trust-domains/{trustDomainName}:
+    get:
+      operationId: GetTrustDomain
+      summary: Retrieves a Trust Domain by name
+      tags:
+        - Trust Domain
+      parameters:
+        - name: trustDomainName
+          description: Trust Domain name
+          schema:
+            $ref: '#/components/schemas/TrustDomainName'
+          in: path
+          required: true
+      responses:
+        '200':
+          description: Successful operation. Returns the Trust Domain
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TrustDomain'
+        default:
+          $ref: '#/components/responses/Default'
+    put:
+      operationId: UpdateTrustDomain
+      summary: Updates a Trust Domain
+      tags:
+        - Trust Domain
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TrustDomainRequest'
+        required: false
+      parameters:
+        - name: trustDomainName
+          description: Trust Domain name
+          schema:
+            $ref: '#/components/schemas/TrustDomainName'
+          in: path
+          required: true
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TrustDomain'
+          description: Successful operation. Returns the updated Trust Domain
+        default:
+          $ref: '#/components/responses/Default'
+    delete:
+      operationId: DeleteTrustDomain
+      summary: Deletes a Trust Domain
+      tags:
+        - Trust Domain
+      parameters:
+        - name: trustDomainName
+          description: Trust Domain name
+          schema:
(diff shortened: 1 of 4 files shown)


pkg/ca/api/harvester.yaml
```diff
@@ -0,0 +1,181 @@
...
+          format: jwt
+          example: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
+      additionalProperties: false
+    Onboard:
+      description: Harvester onboarding data
+      type: object
+      properties:
+        trust_domain:
+          $ref: '#/components/schemas/TrustDomain'
+        # TODO: do we want to include them now?
```
mchurichi, line 111: There are a few open questions about these two fields:
- do we want to include them now? I would say no, it will only add noise to the APIs, and we won't be using them in the near term
- how do we keep trust bundles updated on rotation? we were not planning to send bundle updates to the GCA, should we? do we need answers now or can we just postpone this?

Victorblsilveira: - I would say to remove it until it presents to be necessary.
 - If we decide to send a trust bundle, we would need another pooling mechanism. 
   Similar to what we have between GH <> GS. 
 - Having the bundle updated in the GCA could be good if we want to double-check the bundles. By that, I mean that GS could contact GCA for validation against what it has also, but that seems overkill. (I am not seeing a good use case yet.)
 - Postpone seems like a good option for me.

pkg/ca/api/admin.yaml
```diff
@@ -0,0 +1,262 @@
+openapi: 3.0.3
+info:
+  title: Galadriel Server Admin API - OpenAPI 3.0
+  version: 1.0.11
+  description: Galadriel Server Management API specification.
```
maxlambrecht, line 5: Galadriel CA

pkg/ca/api/admin.yaml
```diff
@@ -0,0 +1,262 @@
+openapi: 3.0.3
+info:
+  title: Galadriel Server Admin API - OpenAPI 3.0
```
maxlambrecht, line 3: Galadriel CA

pkg/ca/api/admin.yaml
```diff
@@ -0,0 +1,262 @@
+openapi: 3.0.3
```
maxlambrecht, line 1: This file would go in `pkg/gca/api`

pkg/ca/api/admin.yaml
```diff
@@ -0,0 +1,262 @@
...
+      operationId: UpdateTrustDomain
+      summary: Updates a Trust Domain
+      tags:
+        - Trust Domain
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TrustDomainRequest'
+        required: false
```
maxlambrecht, line 84: Why it's not required?

pkg/ca/api/harvester.yaml
```diff
@@ -0,0 +1,181 @@
...
+          description: Successful operation. Returns a new access token to be used for authenticating the Trust Domain
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AccessToken'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - join_token: []
+  /trust-domains/access-token:
```
maxlambrecht, line 38: I guess this spec misses the security section with `harvester_auth` requiring an Access Token

pkg/ca/api/harvester.yaml
```diff
@@ -0,0 +1,181 @@
...
+      responses:
+        '201':
+          description: Successful operation. Returns a new access token to be used for authenticating the Trust Domain
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AccessToken'
+        default:
+          $ref: '#/components/responses/Default'
+  /trust-domains/signing-certificates:
```
maxlambrecht, line 53: this one too misses the security section with `harvester_auth` requiring an Access Token

pkg/ca/api/harvester.yaml
```diff
@@ -0,0 +1,181 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AccessToken'
+        default:
+          $ref: '#/components/responses/Default'
+      security:
+        - join_token: []
+  /trust-domains/access-token:
+    post:
```
maxlambrecht, line 39: No need to send a body in this request, thus a `GET` makes more sense here. What is sent by the Harvester is its current access token in the Header `Authorization: Bearer $token`.

pkg/ca/api/harvester.yaml
maxlambrecht: This path is in plural while the previous one (`/trust-domains/access-token`) is in singular.

pkg/ca/api/harvester.yaml
mchurichi: Alright. Removing it for now.

pkg/ca/api/admin.yaml
maxlambrecht: I don't like `gserver` nor `gharvester`, but I think `gca` reads and sounds good, or maybe we just got used to it. Anyway I don't have a strong opinion on this, `ca` or `gca`, as long as we don't add a `g` to everything, that is.

pkg/ca/api/harvester.yaml
wibarre: - Postpone seems like a good option for what we have in the current scope. We will add those when the timing is right.

pkg/ca/api/harvester.yaml
maxlambrecht: Oh missed it. Thanks

wibarre (commented):
Thank you Maxi for putting this. We are very close. Not much additional here.

pkg/ca/api/admin.yaml
```diff
@@ -0,0 +1,262 @@
...
+      responses:
+        '201':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TrustDomain'
+          description: Trust Domain successfully created. Returns newly created Trust Domain
+        default:
+          $ref: '#/components/responses/Default'
+      operationId: CreateTrustDomain
```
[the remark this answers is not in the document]
mchurichi: Definitely 💯

pkg/ca/api/admin.yaml
mchurichi: Sounds reasonable. Moved to `gca/`

mchurichi:
Closing this for now.

PR #124, opened 2023-03-17
Update Server DB models

maxlambrecht:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**
Update Server DB models, specifically removing the fields `digest` and `digest_algorithms`, and renaming the field `signing_cert` to `signing_certificate`.

Fix `Makefile` to run the `sqlc` target on the Server. 


maxlambrecht:
The `digest` in the DB model is not strictly necessary as it can be calculated on the Server, BUT if when implementing the API handlers you realize it's important to have the bundle digests in the DB for performance/efficiency/whatever reasons, add it. Consider that the bundle digests could also be in a cache in the Server, so they would be easily accessible when handling the `BundleSync` requests. 

@Victorblsilveira @jufantozzi @mchurichi

PR #138, opened 2023-04-11
Refactoring Models to use OpenAPI definitions

Victorblsilveira:
**Pull request checklist**

- [x] Proper tests/regressions included?
- [ ] Documentation updated?

**Affected functionality**

Changing handlers to be compliance with API models

**Description of change**

 - Creating abstractions for translating API models into DB models
 - Abstractions for handling conversions  JS <-> bytes for responses and business logic
 - Creating unit tests for conversions 

**Which issue this pull request fix**

Changing the code to be compliance with API specifications


pkg/server/endpoints/management.go
```diff
@@ -8,30 +8,34 @@ import (
...
 
 	"github.com/HewlettPackard/galadriel/pkg/common/util"
 )
 
 func (e *Endpoints) createTrustDomainHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
-	body, err := io.ReadAll(r.Body)
+	var trustDomainReq api.TrustDomain
+	err := fromJSBody(r, &trustDomainReq)
```
mchurichi, line 24: I think we should leverage the structs that the codegen creates. No need to be parsing json in and out for ourselves.

pkg/server/endpoints/management.go
```diff
@@ -278,6 +254,84 @@ func (e *Endpoints) validateToken(ctx echo.Context, token string) (bool, error)
...
+		ID:                uuid,
+		Name:              tdName,
+		CreatedAt:         td.CreatedAt,
+		UpdatedAt:         td.UpdatedAt,
+		Description:       description,
+		OnboardingBundle:  onboardingBundle,
+		HarvesterSpiffeID: harvesterSpiffeID,
+	}, nil
+}
+
```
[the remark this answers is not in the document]
Victorblsilveira: I will work on that, but as a hint, in that case, we may end up just transforming in some well-input defined by the DB API. In other words, in that case, where we are putting bundles, we may end up translating them into a database parameter for the given operation.

maxlambrecht:
My take is that if we follow this approach we will be doing more work than necessary translating between three types of entities/models; I think for this flow API -> DB we need only two:
 
- API entities defined in `common/api` auto-generated from the `schemas.yaml`
- DB models auto-generated from the SQL schemas.

The translation between the two can be encapsulated in the Datastore, thus we need to migrate the current Datastore interface to take in API entities, e.g.:

```
type Datastore interface {
	CreateOrUpdateTrustDomain(ctx context.Context, req *api.TrustDomain) (*api.TrustDomain, error)
```

The current entities in package `entity` should be deleted after the migration. 

PS: one change that we need to make to the `schemas.yaml`: the fields `id` should no be required, as you may no have one if you are creating a new entitiy. The field in the entity should be generated as 
```
Id                *UUID           `json:"id,omitempty"`
```
That way, we can compare the `id` field against `nil` for emptiness in the `CreateOrUpdate` methods.

mchurichi:
I think I know where you are going, and it makes sene, but not sure we can use the autogenerated models the way you suggest as they are.

IMO the entities generated from the schemas are not intended to be plain data models to be used across layers, but mere interfaces for API consumers. They are shaped by the request and response bodies needs, not from an abstract point of view (e.g. some of them lack of IDs, passed as path params, or of any form of unique identifier). You can see them as the autogenerated types from a grpc protobuf, not the "common" types used in spire.

In the interface you proposed, changes to the API spec (upper layer) will have impact on the Datastore interface (lower layer), we don't want that, the dependencies should be top-down for better decoupling, not the other way around.

We can make them serve the purpose of being these "common" data model but we'll need to tweak the spec.

Does it make sense?

maxlambrecht:
> 
> IMO the entities generated from the schemas are not intended to be plain data models to be used across layers, but mere interfaces for API consumers. They are shaped by the request and response bodies needs, not from an abstract point of view (e.g. some of them lack of IDs, passed as path params, or of any form of unique identifier). You can see them as the autogenerated types from a grpc protobuf, not the "common" types used in spire.
>
Well, in SPIRE the common types are generated from protos and are the inputs to the Datastore layer.  

> In the interface you proposed, changes to the API spec (upper layer) will have impact on the Datastore interface (lower layer), we don't want that, the dependencies should be top-down for better decoupling, not the other way around.
> 
> We can make them serve the purpose of being these "common" data model but we'll need to tweak the spec.
> 
> Does it make sense?

It does make sense, it can introduce tight coupling between your API and your data storage layer. 
Then we would have the plain `api.Entities` and a "business layer" with our current entities, which use strong types like `spiffeid.TrustDomain`, etc, and these latter would be the input to the Datastore layer, right?

Ok, in that case, what we need is to add methods to these "business layer" entities to convert from/to `api` entities. We don't want to have `translate` methods scattered across the code. It would be similar to the helpers in the datastore layer:

```
func (r Relationship) ToEntity() (*entity.Relationship, error)
```
The DB model `Relationship` is converted to an entity Relationship, and we could have methods `ToEntity` and `FromEntity`.

mchurichi:
> Well, in SPIRE the common types are generated from protos and are the inputs to the Datastore layer.

Oh nice, I was under the impression that these common types were API specific, but I was wrong. I guess they were designed to be these business models from the very beginning and are simple enough to not be tied to any layer.

> Ok, in that case, what we need is to add methods to these "business layer" entities to convert from/to `api` entities. We don't want to have `translate` methods scattered across the code. It would be similar to the helpers in the datastore layer:
> 
> ```
> func (r Relationship) ToEntity() (*entity.Relationship, error)
> ```
> 
> The DB model `Relationship` is converted to an entity Relationship, and we could have methods `ToEntity` and `FromEntity`.

Yes, that's what we have in mind, regardless of the specific detail implementation the idea is to have some sort of helpers to translate to and from business logic models. I have this simple PoC similar to what Victor is working on that I believe it's somewhat aligned with your proposal: https://github.com/HewlettPackard/galadriel/compare/main...mchurichi:Galadriel:api-test?diff=split

wibarre:
> > Well, in SPIRE the common types are generated from protos and are the inputs to the Datastore layer.
> 
> Oh nice, I was under the impression that these common types were API specific, but I was wrong. I guess they were designed to be these business models from the very beginning and are simple enough to not be tied to any layer.
> 
> > Ok, in that case, what we need is to add methods to these "business layer" entities to convert from/to `api` entities. We don't want to have `translate` methods scattered across the code. It would be similar to the helpers in the datastore layer:
> > ```
> > func (r Relationship) ToEntity() (*entity.Relationship, error)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The DB model `Relationship` is converted to an entity Relationship, and we could have methods `ToEntity` and `FromEntity`.
> 
> Yes, that's what we have in mind, regardless of the specific detail implementation the idea is to have some sort of helpers to translate to and from business logic models. I have this simple PoC similar to what Victor is working on that I believe it's somewhat aligned with your proposal: https://github.com/HewlettPackard/galadriel/compare/main...mchurichi:Galadriel:api-test?diff=split

I had similar thoughts for the GCA. Decoupling the translation from the datastore makes a lot of sense. We could apply a similar approach when time is right for the GCA.

pkg/server/endpoints/run.go
```diff
@@ -127,7 +129,17 @@ func (e *Endpoints) addHandlers() {
 }
 
 func (e *Endpoints) addTCPHandlers(server *echo.Echo) {
-	server.CONNECT("/onboard", e.onboardHandler)
-	server.POST("/bundle", e.postBundleHandler)
-	server.POST("/bundle/sync", e.syncFederatedBundleHandler)
+	sw := harvesterAPI.ServerInterfaceWrapper{
+		Handler: e,
+	}
```
mchurichi, line 134: What's the benefit of defining this wrapper? I think we can simply implement this interface in a new Handler struct. That way we don't need to re-define the paths and handlers one by one as below.
See https://github.com/HewlettPackard/galadriel/compare/main...mchurichi:Galadriel:api-test?diff=split#diff-05cbb367d3dd1e9edd60959c6b72e59497c7192d07993dc02c703583fa6abb35R132

mchurichi: And at the same time we can keep separated the Handler implementation and listeners setup

maxlambrecht: I agree, we don't need the wrapper, but I think it's a good idea to have handlers in their own packages and structs, as an example we can define the handler like this: https://github.com/HewlettPackard/galadriel/blob/48e74658edcb129364d24a18e95ba092fd74a8c8/pkg/gca/endpoints/jwt/handler.go#L46

And then add the handler to the Mux:

```
        // TCP API handlers
	mux := http.NewServeMux()

	jwtHandler, err := jwt.NewHandler(&jwt.Config{
		CA:          e.CA,
		Logger:      e.Logger,
		JWTTokenTTL: e.config.JWTTokenTTL,
		Clock:       e.Clock,
	})
	if err != nil {
		return fmt.Errorf("failed to build JWT handler")
	}
	mux.Handle("/jwt", jwtHandler)
```

Victorblsilveira:
@maxlambrecht @mchurichi @wibarre @jufantozzi I recently pushed some changes that affect the project structure somehow. Please take a look there, I would love some feedback.
- New HTTP package for generic used case functions -(parsing request and response bodies and error handling) for echo and for standard golang HTTP package
- Helper.go files close to the .gen.go definitions to inject in the structures the ToEntity function for converting API models into business models.
Let me know what you all think.

pkg/common/http/http.go
```diff
@@ -0,0 +1,98 @@
...
+	return nil
+}
+
+/*
+FromJSBody parses json bytes into a struct
+r: Request that contains the json bytes to be parsed into 'in'
+in: Reference(pointer) to the interface to be full filled
+*/
+func FromJSBody(r *http.Request, in interface{}) error {
+	body, err := io.ReadAll(r.Body)
```
mchurichi, line 84: Let's use generics or a custom interface here. This is very likely to cause problems.

pkg/common/http/http.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil
+	}
+
+	return write(ctx, body)
+}
+
+/*
+WriteArrayResponse parses a slice of object into a json and writes in the response
+*/
+func WriteArrayResponse[T any](ctx echo.Context, body []*T) error {
```
mchurichi, line 36: After the last refactor I don't think we need these to use generics anymore, as everything ends up being used as interface{} when sent as a parameter to ctx.JSON. Also, we have two functions that do exactly the same. Can switch back to a single function taking an interface?

pkg/server/endpoints/auth.go
```diff
@@ -0,0 +1,30 @@
...
+	Datastore datastore.Datastore
+	Logger    logrus.FieldLogger
+}
+
+func (m AuthenthicationMD) Authenticate(token string, ctx echo.Context) (bool, error) {
+
+	gctx := ctx.Request().Context()
+
+	// Any skip cases ?
+	t, err := m.Datastore.FindJoinToken(gctx, token)
```
mchurichi, line 19: If the Join Token is not found in the database the datastore returns `nil, nil` (which is correct) and causes the log call to panic when reading `t.TrustDomainID`.
We need to check for a nil `t` and handle this case properly.

pkg/server/endpoints/harvester.go
```diff
@@ -1,258 +1,134 @@
...
+	chttp "github.com/HewlettPackard/galadriel/pkg/common/http"
 	"github.com/HewlettPackard/galadriel/pkg/common/util"
-	"github.com/google/uuid"
+
 	"github.com/labstack/echo/v4"
-	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/sirupsen/logrus"
+
+	commonAPI "github.com/HewlettPackard/galadriel/pkg/common/api"
+	harvesterAPI "github.com/HewlettPackard/galadriel/pkg/server/api/harvester"
```
maxlambrecht, line 16: why not use `harvester` package?

pkg/server/endpoints/harvester.go
```diff
@@ -1,258 +1,134 @@
...
-func (e *Endpoints) syncFederatedBundleHandler(ctx echo.Context) error {
-	e.Logger.Debug("Receiving sync request")
+// Upload a new trust bundle to the server
+// (PUT /trust-domain/{trustDomainName}/bundles)
+func (h HarvesterAPIHandlers) BundlePut(ctx echo.Context, trustDomainName commonAPI.TrustDomainName) error {
+	h.Logger.Debug("Receiving post bundle request")
+	gctx := ctx.Request().Context()
 
+	// TODO: move authn out and replace with Access Token when implemented
 	jt, ok := ctx.Get(tokenKey).(*entity.JoinToken)
```
maxlambrecht, line 65: isn't this part of the Auth middleware?

pkg/server/endpoints/auth.go
```diff
@@ -0,0 +1,43 @@
+package endpoints
```
maxlambrecht, line 1: missing tests.

pkg/server/endpoints/harvester.go
```diff
@@ -1,258 +1,134 @@
 package endpoints
```
maxlambrecht, line 1: missing tests

pkg/server/endpoints/management.go
```diff
@@ -2,196 +2,170 @@ package endpoints
...
-	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
-	"time"
 
+	commonAPI "github.com/HewlettPackard/galadriel/pkg/common/api"
 	"github.com/HewlettPackard/galadriel/pkg/common/entity"
+	chttp "github.com/HewlettPackard/galadriel/pkg/common/http"
+	adminAPI "github.com/HewlettPackard/galadriel/pkg/server/api/admin"
```
maxlambrecht, line 11: no need to rename the import package

pkg/server/endpoints/management.go
```diff
@@ -2,196 +2,170 @@ package endpoints
```
maxlambrecht, line 2: missing tests

pkg/common/api/helper.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (td TrustDomain) ToEntity() (*entity.TrustDomain, error) {
+	harvesterSpiffeID, err := spiffeid.FromString(*td.HarvesterSpiffeId)
+	if err != nil {
+		return nil, common.ErrWrongSPIFFEID{Cause: err}
+	}
+
+	tdName, err := spiffeid.TrustDomainFromString(td.Name)
+	if err != nil {
+		return nil, common.ErrWrongTrustDomain{Cause: err}
```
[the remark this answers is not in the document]
Victorblsilveira: The first point is to be reusable, and we could use the other approach declaring the error at the beginning of the file as a variable. 
The second point is, since we are interpolating the error messages, the usual approach of defining it as a var is less practical than this one since we would have to interpolate every place that would be used. So I decided to centralize in that way.

pkg/common/api/helper.go
```diff
@@ -0,0 +1,72 @@
...
+	return &Relationship{
+		Id:                  entity.ID.UUID,
+		CreatedAt:           entity.CreatedAt,
+		UpdatedAt:           entity.UpdatedAt,
+		TrustDomainAId:      entity.TrustDomainAID,
+		TrustDomainBId:      entity.TrustDomainBID,
+		TrustDomainBConsent: entity.TrustDomainBConsent,
+		TrustDomainAConsent: entity.TrustDomainAConsent,
+	}
+}
```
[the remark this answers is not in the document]
Victorblsilveira: We may need it, but as I was talking in one of our dailies, some flows will be hard to predict and I prefer to implement that conversion when the time comes. My next activity that will address the implementation of the handlers will be a good fit.

pkg/server/endpoints/auth.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (m AuthenthicationMD) Authenticate(token string, ctx echo.Context) (bool, error) {
+	gctx := ctx.Request().Context()
+
+	// Any skip cases ?
+	t, err := m.Datastore.FindJoinToken(gctx, token)
+	if err != nil {
+		m.Logger.Errorf("Invalid Token: %s\n", token)
+		message := "Invalid authorization token"
+		return false, echo.NewHTTPError(http.StatusUnauthorized, message)
```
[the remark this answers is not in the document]
Victorblsilveira: Well, this flow is already implemented. I just moved it, I think we will need to revisit this for sure, but my intent here was just to centralize and make this behavior clear.

pkg/server/endpoints/harvester.go
maxlambrecht: Ok, this breeds a question: why did you add the Auth middleware in a PR for refactoring API entities?

pkg/server/endpoints/management.go
maxlambrecht: Add the `_test` files with a `// TODO: missing tests will be added when the API are reimplemented`, so when you implement the API you take care of that.

pkg/common/api/helper.go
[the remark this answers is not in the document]
Victorblsilveira: The value that I see is really the reusability and avoiding doing what you mention. Calling `fmt.Errorf("...: %w")` can lead to mistyping in many places, and in case of error refactoring things get worse. I am ok in cases where the error is only used in one place, which is not the case here. Let's see what the others think, but I prefer the first approach.

pkg/server/endpoints/auth.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	if t == nil {
+		message := "Token not found"
+		return false, echo.NewHTTPError(http.StatusForbidden, message)
+	}
+
+	m.Logger.Debugf("Token valid for trust domain: %s\n", t.TrustDomainID)
+	ctx.Set("token", t)
```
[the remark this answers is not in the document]
Victorblsilveira: BundlePut is using this token.

pkg/server/endpoints/harvester.go
Victorblsilveira: I already did the change to use `harvester` but I think that is kind of misleading since it is and API, and the package name does make it clear when importing. That was the reason behind the package renaming at import.

pkg/server/endpoints/harvester.go
Victorblsilveira: I mention that I am changing the code to be more compliant with the echo standard. That includes the middleware, if you see the old auth implementation I was trying to not change the behavior.

pkg/server/endpoints/management.go
maxlambrecht: but if you need to rename packages, use lowercase: `adminapi`

pkg/common/api/helper.go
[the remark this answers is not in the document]
maxlambrecht: I wouldn't expect many places where the trust domain is parsed, I think it will be in the `ToEntity` methods.

Also the returned error should say which was the trust domain that failed to be parsed:

```
td, err := spiffeid.TrustDomainFromString(td.Name)
if err != nil {
	return nil, fmt.Errorf("failed to parse trust domain %q: %w", td.Name, err)
}
```
Other thing, the variable for the parsed trust domain should be `td` or `trustDomain`.

I agree that it's a good idea to get input from the rest of the team.

maxlambrecht:
@Victorblsilveira, there is a failing check, the PR build.

Victorblsilveira:
> @Victorblsilveira, there is a failing check, the PR build.

Yeah, it's propositional to remind me of some unit tests I need to implement. 
I am taking care of them right now.

pkg/common/api/helper.go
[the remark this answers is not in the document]
mgbcaio: My opinion on this is that I prefer going with the normal way. It makes the code easier to maintain and more readable and easy interpretation to find the failure point of the function, although your method looks way more organized and less verbose :). 

But going the common way, even tho we might need to interpolate strings contstantly, if we ever need to change the behavior of a function it would be much simpler to just change the string that goes along with the error message, and not create a new field, or change it in a another file.

That’s what I think about it.

pkg/server/datastore/fakedatabase.go
```diff
@@ -0,0 +1,457 @@
...
+	db.mutex.Lock()
+	defer db.mutex.Unlock()
+
+	if err := db.getNextError(); err != nil {
+		return nil, err
+	}
+
+	relationships := []*entity.Relationship{}
+	for _, r := range db.relationships {
+		mactchA := r.TrustDomainAID.String() == trustDomainID.String()
```
maxlambrecht, line 388:
```suggestion
		matchA := r.TrustDomainAID.String() == trustDomainID.String()
```

pkg/server/datastore/fakedatabase.go
```diff
@@ -0,0 +1,457 @@
...
+		mattchB := r.TrustDomainBID.String() == trustDomainID.String()
```
maxlambrecht, line 389:
```suggestion
		matchB := r.TrustDomainBID.String() == trustDomainID.String()
```

pkg/server/datastore/fakedatabase.go
```diff
@@ -0,0 +1,457 @@
...
+
+		if mactchA || mattchB {
```
maxlambrecht, line 391:
```suggestion
		if matchA || matchB {
```

pkg/server/datastore/fakedatabase.go
```diff
@@ -0,0 +1,457 @@
...
+
+	// Entities
+	bundles       map[uuid.UUID]*entity.Bundle
+	tokens        map[uuid.UUID]*entity.JoinToken
+	trustDomains  map[uuid.UUID]*entity.TrustDomain
+	relationships map[uuid.UUID]*entity.Relationship
+}
+
+func NewFakeDB() *FakeDatabase {
+
```
maxlambrecht, line 25: remove line

pkg/server/endpoints/run.go
```diff
@@ -4,13 +4,17 @@ import (
...
 
 	"github.com/HewlettPackard/galadriel/pkg/common/util"
 	"github.com/HewlettPackard/galadriel/pkg/server/datastore"
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"github.com/sirupsen/logrus"
+
+	"github.com/HewlettPackard/galadriel/pkg/common/telemetry"
+
+	adminAPI "github.com/HewlettPackard/galadriel/pkg/server/api/admin"
```
maxlambrecht, line 16:
```suggestion
	adminapi "github.com/HewlettPackard/galadriel/pkg/server/api/admin"
```

When renaming an imported package, the local name should follow the same guidelines as package names (lower case, no under_scores or mixedCaps).

pkg/server/endpoints/run.go
```diff
@@ -4,13 +4,17 @@ import (
...
+	harvesterAPI "github.com/HewlettPackard/galadriel/pkg/server/api/harvester"
```
maxlambrecht, line 17:
```suggestion
	harvesterapi "github.com/HewlettPackard/galadriel/pkg/server/api/harvester"
```

pkg/server/api/admin/helper_test.go
```diff
@@ -0,0 +1,54 @@
...
+	t.Run("Does not allow wrong trust domain names", func(t *testing.T) {
+		description := "a cool description"
+
+		tdPut := TrustDomainPut{
+			Name:        "A wrong trust domain name",
+			Description: &description,
+		}
+
+		trustDomain, err := tdPut.ToEntity()
+		assert.Error(t, err)
```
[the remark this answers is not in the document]
Victorblsilveira: How do I suppose to do that since the error message is being interpolated? Suggestion?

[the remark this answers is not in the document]
mchurichi: we know the result of the error message interpolation beforehand, and we can assert the error message to start with something like:
`malformed trust domain[A wrong trust domain name]:`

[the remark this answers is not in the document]
Victorblsilveira: I used `ErrorContains` asserting the `malformed trust domain`. Is it enough?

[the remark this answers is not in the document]
mchurichi: I'd have the trust domain checked too, we already know the value and that way we make sure we are using the right value

maxlambrecht (dismissed):
Thanks for this work, @Victorblsilveira!

mchurichi (approved):
🍨

maxlambrecht (approved):

mchurichi (approved):

PR #151, opened 2023-05-08
Add JWT issuing and validation. Change Harvester onboarding and authentication processes.

maxlambrecht:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**

Changed the harvester onboarding and authentication process:

- The onboarding process uses a `join_token`, validates it's still valid and not used, and returns a JWT token with the subject equal to the onboarded trust domain name.

- The authentication of API calls is done using a JWT validator to verify the Bearer token from the Authorization header. 

**Description of change**

- Added interfaces `jwt.Issuer`, `jwt.Validator`, and `KeyManager`, to issue and validate JWT tokens, using keys managed by a `KeyManager`. 

- The implementation added for the `KeyManager` interface is a simple in-memory manager.

- Added a goroutine that request a new JWT token every 12 hours.  

- Added decorator to every API call in the Harvester to add the JWT Bearer token.

Fixes [GLCP-45111](https://hpe.atlassian.net/browse/GLCP-45111)


maxlambrecht:
WIP: unit tests for the `Onboard` and `GetNewJWTToken`

pkg/common/jwt/issuer_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func TestNew(t *testing.T) {
+	signer, err := cryptoutil.GenerateSigner(cryptoutil.RSA2048)
+	require.NoError(t, err)
+
+	ca, err := NewJWTCA(&Config{
+		Signer: signer,
+		Kid:    "test-kid",
+	})
+	require.NoError(t, err)
```
Victorblsilveira, line 23: assert error message?

pkg/common/jwt/issuer_test.go
```diff
@@ -0,0 +1,85 @@
...
+	require.NotNil(t, ca)
+	assert.Equal(t, "test-kid", ca.kid)
+	assert.Equal(t, signer, ca.signer)
+	assert.NotNil(t, ca.clk)
+
+	// missing signer
+	ca, err = NewJWTCA(&Config{
+		Kid: "test-kid",
+	})
+	require.Error(t, err)
```
Victorblsilveira, line 33: same here

pkg/common/jwt/issuer_test.go
```diff
@@ -0,0 +1,85 @@
...
+	require.Nil(t, ca)
+	assert.Equal(t, "signer is required", err.Error())
+
+	// missing kid
+	ca, err = NewJWTCA(&Config{
+		Signer: signer,
+	})
+	require.Error(t, err)
```
Victorblsilveira, line 41: here too, I think that are more cases. You may need to review error message assertions

pkg/common/jwt/validator.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+	}
+
+	if !validAudience {
+		return nil, jwt.NewValidationError("token audience is invalid", jwt.ValidationErrorAudience)
+	}
+
+	return claims, nil
+}
```
Victorblsilveira, line 66: I think that encapsulating audience verification would make the code simpler. Something similar to this:
```suggestion
func (v *DefaultJWTValidator) ValidateToken(ctx context.Context, token string) (*jwt.RegisteredClaims, error) {
	claims := &jwt.RegisteredClaims{}
	_, err := jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
		return v.getPublicKey(ctx, token)
	})
	if err != nil {
		return nil, fmt.Errorf("failed to parse and validate token: %w", err)
	}

	// Validate the claims
	err = claims.Valid()
	if err != nil {
		return nil, err
	}
	
	if !v.validAudience(claims) {
		return nil, jwt.NewValidationError("token audience is invalid", jwt.ValidationErrorAudience)
	}

	return claims, nil
}

func (v *DefaultJWTValidator) validAudience(claims *jwt.RegisteredClaims) bool {
	for _, aud := range v.expectedAudience {
		ok := claims.VerifyAudience(aud, true)
		if !ok {
			return false
		}
	}
	return true
}
```

pkg/harvester/client/server.go
```diff
@@ -23,63 +26,127 @@ const (
...
+		return errors.New("empty response body")
+	}
+	c.jwtProvider.setToken(bodyString)
+
 	c.logger.Info("Connected to Galadriel Server")
+
+	return nil
+}
+
+func (c *client) GetNewJWTToken(ctx context.Context) error {
```
Victorblsilveira, line 117: When this is supposed to be used?

pkg/harvester/client/server.go
```diff
@@ -172,18 +229,61 @@ func createTLSClient(trustBundlePath string) (*http.Client, error) {
...
+
+// RoundTrip applies the decorator to every request adding the Authorization header
+func (t *decoratedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.skipper != nil && t.skipper(req) {
+		return t.transport.RoundTrip(req)
+	}
+
+	// Apply decorator
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", t.jwtProvider.getToken()))
+	req.Header.Set("Content-Type", contentType)
```
Victorblsilveira, line 262: Should we rename `contentType` to a real content type name like `jsonContentType` or something similar? Is really not clear here what is the content type being used.

pkg/server/endpoints/harvester.go
```diff
@@ -4,98 +4,187 @@ import (
...
+		TTL:      24 * 5 * time.Hour,
+	}
+
+	jwtToken, err := h.jwtIssuer.IssueJWT(ctx, jwtParams)
+	if err != nil {
+		msg := fmt.Errorf("error generating JWT token")
+		return h.handleErrorAndLog(err, msg, http.StatusInternalServerError)
+	}
+
+	return echoCtx.JSON(http.StatusOK, jwtToken)
```
Victorblsilveira, line 110: You should use the wrapped methods in our http package. Something like: 
```suggestion
	return chttp.WriteResponse(echoCtx, jwtToken)
```

Victorblsilveira (commented):
Excellent work here @maxlambrecht really good changes. I have just some minor doubts.|
See if make sense.

pkg/common/jwt/issuer.go
```diff
@@ -0,0 +1,88 @@
...
+
+	clk clock.Clock
+}
+
+// NewJWTCA creates a new JWTCA.
+func NewJWTCA(c *Config) (*JWTCA, error) {
+	if c.Signer == nil {
+		return nil, fmt.Errorf("signer is required")
+	}
+	if c.Kid == "" {
```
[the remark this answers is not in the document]
maxlambrecht: No, it's not optional. The `kid` should be present to set it in the JWT header to be able to associate it to a public key for validation.

pkg/common/jwt/issuer_test.go
maxlambrecht: It's asserting that there shouldn't be any error.

pkg/common/jwt/issuer_test.go
maxlambrecht: The assert for the error message is 2 lines after.

pkg/common/jwt/issuer_test.go
maxlambrecht: It's asserting that there was an error, as expected, and 2 lines below there's an assertion for the error message.

pkg/harvester/client/server.go
maxlambrecht: It's being used in the goroutine that keeps requesting a new JWT token every few hours, the JWT token rotator.

pkg/common/jwt/issuer_test.go
maxlambrecht: or should I replace it by `assert.NoError`?

Victorblsilveira: that's good, my comment doest make sense, sorry.

Victorblsilveira (dismissed):
I left 1 nit comment. So feel free to skip it if necessary. 
All good, good job @maxlambrecht

maxlambrecht:
> I left 1 nit comment. So feel free to skip it if necessary. All good, good job @maxlambrecht

Thanks @Victorblsilveira. 
I added tests for the Onboard handler, and I'm adding tests for the GetJWT handler.

maxlambrecht:
@Victorblsilveira I refactored the Onboard and GetNewJWTToken requests using a method `callAPI` to remove duplication. The rest the the harvester request method could use the same.

Victorblsilveira:
> @Victorblsilveira I refactored the Onboard and GetNewJWTToken requests using a method `callAPI` to remove duplication. The rest the the harvester request method could use the same.

I just realized that yesterday, on my bed, when trying to sleep. Why don't we use the client boilerplate for making that calls?

Victorblsilveira (approved):
Really good, there are some minors, like, Can we use the client boilerplates to consume Galadriel API? But I really think that you can decide whether to use it or not. Besides that, I think that your PR is good to go. +1

pkg/server/endpoints/harvester.go
```diff
@@ -4,98 +4,193 @@ import (
...
 	}
-	// end authn
 
 	req := &harvester.BundlePutJSONRequestBody{}
-	err = chttp.FromBody(ctx, req)
+	err := chttp.FromBody(ctx, req)
 	if err != nil {
 		err := fmt.Errorf("failed to read bundle put body: %v", err)
-		return h.handleErrorAndLog(err, http.StatusBadRequest)
+		return h.handleErrorAndLog(err, err, http.StatusBadRequest)
```
mgbcaio, line 183: Why do we need to pass the `err` multiple times, cant the function use the same parameter  ?

mgbcaio (commented):
just 2 comments, see if it makes sense =) Great job, Max. 🦕

pkg/server/endpoints/harvester.go
maxlambrecht: Because the method both logs the error and handles the error sending back a message to the caller, thus it takes in an error to log and an error to return, that usually are not the same.

mgbcaio (approved):
🍦

PR #158, opened 2023-05-16
Consolidate functions for filtering and mapping relationships.

maxlambrecht:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**
N/A

**Description of change**
Refactor common functionality for management and harvester API. 
Refactor and improve tests. 


pkg/server/endpoints/management.go
```diff
@@ -363,65 +370,6 @@ func (h *AdminAPIHandlers) lookupTrustDomain(ctx context.Context, trustDomainID
...
-	return ca == api.Denied || cb == api.Denied
-}
-
-func accepted(ca api.ConsentStatus, cb api.ConsentStatus) bool {
-	return ca == api.Accepted && cb == api.Accepted
-}
-
-func pending(ca api.ConsentStatus, cb api.ConsentStatus) bool {
-	return (ca == api.Pending || cb == api.Pending) && ca != api.Denied && cb != api.Denied
-}
```
Victorblsilveira, line 412: The `FilterRelationships` can handle these cases? I mean, 

From the Admin perspective:
- The relation is `Accepted` if both accept.
- The relation is `Pending` if no one was denied and at least one is pending.
- The relation is `Denied` if at least one is denied.

From my understanding the FilterRelationships is checking both statuses, so we will miss the pending and denied behaviors.

Victorblsilveira (commented):
Pretty good so far. Just one doubt.

pkg/server/endpoints/management.go
maxlambrecht: The relationships status don't have that semantics. A relationship entity in reality is representing two uni-directional relationships
When an Admin creates a relationship between TD-A and TD-B, that relationship has two `pending` status. When TD-A accepts, it changes to  `accepted` - `pending`, so it's an accepted relationship for one of the trust domains, so TD-A will start getting bundles from TD-B. Then TD-B may accept or deny or do nothing, and what it does will change its side of the relationship.

Victorblsilveira: if trust domain A accepts the relationship and Trust Domain B denies it, the trust domain A should keep receiving the trust bundle from B?

maxlambrecht: > What the admin will receive if it only filters by pending, active or denied relationships without specifying a trust domain?

Will get all relationships that has the status in either of the trust domain consents:

```
isConsentStatusMatch = trustDomainAConsent == status || trustDomainBConsent == status
```

> if trust domain A accepts the relationship and Trust Domain B denies it, the trust domain A should keep receiving the trust bundle from B?

Yes. TD-A says it wants to accept bundles from TD-B.

Victorblsilveira: > > What the admin will receive if it only filters by pending, active or denied relationships without specifying a trust domain?
> 
> Will get all relationships that has the status in either of the trust domain consents:
> 
> ```
> isConsentStatusMatch = trustDomainAConsent == status || trustDomainBConsent == status
> ```
> 
> > if trust domain A accepts the relationship and Trust Domain B denies it, the trust domain A should keep receiving the trust bundle from B?
> 
> Yes. TD-A says it wants to accept bundles from TD-B.

I don't think that receiving a bundle from a party that denied the relationship makes sense for the federation. What will be the purpose of the bundle?

maxlambrecht: If TD-A accepts a bundle from TD-B, it will be able to validate certificates/identities from TD-B and do TLS.
Denying a relationships means that you won't accept bundles from the other party, but your bundles still are uploaded to Galadriel Server anyway.

Victorblsilveira: It is possible to only do TLS in the SPIRE context?

Victorblsilveira: I mean, can we do the federation using only TLS?

Victorblsilveira: I had in mind that we are going to the mutual consent relationship but seems that we are going also for the partial consent.

maxlambrecht: Regardless, I believe that consolidating the logic into a single, well-documented `FilterRelationships` function is a move in the right direction. It simplifies maintenance and enhances readability.

Victorblsilveira (approved):
+1

pkg/server/endpoints/management.go
Victorblsilveira: > Regardless, I believe that consolidating the logic into a single, well-documented `FilterRelationships` function is a move in the right direction. It simplifies maintenance and enhances readability.

Totally agree, I was just in doubt of the meanings, you got my +1

PR #165, opened 2023-05-18
Add Digest to Bundle models

maxlambrecht:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**
N/A

**Description of change**
In order to make things simpler in the Server, it's best to store the bundle digests and also to require the Harvester to send the Digest as part of the Bundle upload, so the Server doesn't need to do that expensive operation.

Removed `SignatureAlgorithm` field that is not used right now.

Add digest verification on the server. 

Consolidate digest and encoding function as common utility 

Make `Signature` and `Signing_cert` not required to support the scenario where the Bundle verification is opted out.


Victorblsilveira (commented):
The changes are OK, I don't know if we can do that. 
Even if the harvester sends the digest, the server still needs to calculate to ensure the digest is correct.

Victorblsilveira (commented):
The changes are OK, I don't know if we can do that. 
Even if the harvester sends the digest, the server still needs to calculate to ensure the digest is correct.

Victorblsilveira:
> 



> > The changes are OK, I don't know if we can do that. Even if the harvester sends the digest, the server still needs to calculate to ensure the digest is correct.
> 
> The server is already trusting that the Harvester is sending the correct bundle, so it seems reasonable to also trust that the Harvester is sending the correct digest. The server is primarily using the digests for bundle comparison rather than for bundle validation.

Sending the wrong bundle will make the communications fail. Sending the wrong digest will make unnecessary bundle syncs. I know that we trust the harvester but that does mean we can skip validations.

maxlambrecht:
> > 
> 
> > > The changes are OK, I don't know if we can do that. Even if the harvester sends the digest, the server still needs to calculate to ensure the digest is correct.
> > 
> > 
> > The server is already trusting that the Harvester is sending the correct bundle, so it seems reasonable to also trust that the Harvester is sending the correct digest. The server is primarily using the digests for bundle comparison rather than for bundle validation.
> 
> Sending the wrong bundle will make the communications fail. Sending the wrong digest will make unnecessary bundle syncs. I know that we trust the harvester but that does mean we can skip validations.

Ok, I'll add a validation on the Server.

Victorblsilveira (approved):
+1

PR #173, opened 2023-05-26
Improvements in CLI implementations

maxlambrecht:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

This pull request introduces several changes to improve code organization, readability, and testability in the CLI implementation:

1. **HTTP Package Restructuring:**
   - Moved common functionality for reading responses from an HTTP response to the `http` package.
   - Moved the common Unix Domain Socket (UDS) client creation logic to the `http` package.

2. **Constant Flag Names:**
   - Defined constants for CLI flag names, improving code clarity and reducing duplication.

3. **Dependency Injection in NewUDSClient:**
   - Modified the `NewUDSClient` function to accept an `http.Client` dependency, allowing for easier testing and improved flexibility.

These modifications aim to enhance maintainability, readability, and testability of the codebase.


cmd/server/util/client.go
```diff
@@ -34,31 +32,25 @@ type GaladrielAPIClient interface {
...
-		DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
-			var d net.Dialer
-			return d.DialContext(ctx, "unix", socketPath)
-		}}
+	if httpClient == nil {
+		httpClient = httputil.NewUDSHTTPClient(socketPath)
+	}
 
-	client.Client = &http.Client{Transport: t}
+	client.Client = httpClient
```
Victorblsilveira, line 53: Not related to what you have done here, but that is worth thinking about. 
The word client appears many times here with different meanings in each of them. 
- Client in the sense of Golang client
- Client as an API consumer for Galadriel

My suggestion is that we rename it for something that aligns with the purpose.

Maybe changing 
```
	adminAPIClient, err := admin.NewClient(socketPath, adminClientOpt)
	if err != nil {
		return nil, fmt.Errorf("failed to instantiate the Admin Client: %v", err)
	}
```
 and then rename `httpClient` to `client` since it is the common pattern for HTTP clients.
 
 Other names rather than `adminAPIClient`: 
 - `adminAPI`
 - `adminClient`
 - `adminAPIConsumer`
 
 WDYT?

Victorblsilveira (approved):
All good +1

Just one comment that is not mandatory, I leave it up to you to decide whether to address it here or not.

cmd/server/util/client.go
maxlambrecht: I like `adminClient`.

PR #182, opened 2023-05-31
Adding pagination / sorting in API

Victorblsilveira:
**Pull request check list**

- [x] Proper tests/regressions included?
- [x] Documentation updated?

**Affected functionality**

**Description of change**

Adding more flexibility in querying data through API 
- Pagination
- Sorting


Victorblsilveira:
Regarding the dependencies in `startDate` and `endDate` params. There are some limitations with OpenAPI definitions and params dependencies: https://swagger.io/docs/specification/describing-parameters/

We will need to enforce the requirements of both params in the implementation.

pkg/common/api/schemas.yaml
```diff
@@ -180,6 +180,21 @@ components:
       type: string
       description: base64 encoded signature of the bundle
       example: aG6ZX3aZf+6tg/yRDZdRSNtM39lzZWNyZXRJZD1JYW1TZWNyZXRJZCZjdXJyZW50VGltZVN0YW1wPTE1NzEyMTUwOTUmZXhwaXJlVGltZT0xODg2NTc1MDk1JnJhbmRvbT0yMjA2MjU=
+    Offset:
```
maxlambrecht, line 183: Why not use the more intuitive parameters `pageSize` and `pageNumber`?

maxlambrecht:
I'd suggest to focus the PR on pagination for the server admin API,  maybe just for Relationships for example, and to make the necessary changes in the handlers and other related functionality to support the new API. It'll need changes in the Datastore layer as well.

Victorblsilveira:
@maxlambrecht @mgbcaio 
The last commits successfully enable flexible queries with pagination and consent status.
Take a look to see if the approach makes sense. If this is approved, we will need to refactor the whole database queries and that is a lot of work and refactoring. So I want to make sure that this will be the approach that we all want.

pkg/common/jwt/validator.go
```diff
@@ -65,12 +66,20 @@ func (v *DefaultJWTValidator) validAudience(claims *jwt.RegisteredClaims) bool {
 }
 
 func (v *DefaultJWTValidator) getPublicKey(ctx context.Context, t *jwt.Token) (crypto.PublicKey, error) {
-	kid := t.Header[kidHeader].(string)
+	if t != nil {
```
maxlambrecht, line 69: Besides that I don't see any reason to change this method in this PR, the changes don't make much sense to me: 
- `t` will never be nil. 
- the returned error "jwt token not found" is confusing, the method doesn't look up a token. 

The only improvements that could be made are: 
- add context to the returned error 
- add the `ok` in the first line and check it. 

The method would look like this:

```
func (v *DefaultJWTValidator) getPublicKey(ctx context.Context, t *jwt.Token) (crypto.PublicKey, error) {
	kid, ok := t.Header[kidHeader].(string)
	if !ok {
		return nil, fmt.Errorf("failed to retrieve kid from token header")
	}

	key, err := v.keyManager.GetKey(ctx, kid)
	if err != nil {
		return nil, fmt.Errorf("failed to retrieve public key for kid %s: %w", kid, err)
	}

	return key.Signer().Public(), nil
}
```

If you want to make this changes, submit it in separate PR.

Victorblsilveira: Solved here: https://github.com/HewlettPackard/galadriel/pull/192

Victorblsilveira:
Closing because most of the ideas were already implemented in https://github.com/HewlettPackard/galadriel/pull/193 and https://github.com/HewlettPackard/galadriel/pull/192

PR #201, opened 2023-06-06
Fix issue with server `socket_path`

maxlambrecht:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**

Previously, the `galadriel-server run` command did not utilize the `socket_path` property from the config file due to the `socketPath` flag taking precedence over the config file, even with a default value set.

The issue has been resolved by removing the default value for the `socketPath` flag in the run command. As a result, the `galadriel-server` now correctly utilizes the `socket_path` property when executing the run command.

**Description of change**


Victorblsilveira (approved):
+1

PR #224, opened 2023-06-14
Update, Delete and List operations for Relationships CLI

mgbcaio:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**
Relationship server admin CLI

**Description of change**
added the ability to perform the update, deletion and listing of relationships.

[GLCP-39474](https://hpe.atlassian.net/browse/GLCP-39474)


pkg/server/endpoints/admin.go
```diff
@@ -325,6 +325,92 @@ func (h *AdminAPIHandlers) PutTrustDomainByName(echoCtx echo.Context, trustDomai
...
+
+	rel, err := reqBody.ToEntity()
+	if err != nil {
+		err := fmt.Errorf("failed to read relationship patch body: %v", err)
+		return chttp.LogAndRespondWithError(h.Logger, err, err.Error(), http.StatusBadRequest)
+	}
+
+	relDB, err := h.findRelationshipByID(ctx, relationshipID)
+	if err != nil {
+		return err
```
Victorblsilveira, line 347:
```suggestion
		return chttp.LogAndRespondWithError(h.Logger, err, err.Error(), http.StatusInternalServerError)
```

pkg/server/endpoints/admin.go
```diff
@@ -325,6 +325,92 @@ func (h *AdminAPIHandlers) PutTrustDomainByName(echoCtx echo.Context, trustDomai
...
+}
+
+// DeleteRelationshipByID deletes a specific relationship based on its id - (DELETE /relationships/{relationshipID})
+func (h *AdminAPIHandlers) DeleteRelationshipByID(echoCtx echo.Context, relationshipID api.UUID) error {
+	ctx := echoCtx.Request().Context()
+
+	//Check if the relationship exist in the database
+	relationship, err := h.findRelationshipByID(ctx, relationshipID)
+	if err != nil {
+		return err
```
Victorblsilveira, line 391:
```suggestion
		return chttp.LogAndRespondWithError(h.Logger, err, err.Error(), http.StatusInternalServerError)
```

pkg/server/endpoints/admin.go
```diff
@@ -325,6 +325,92 @@ func (h *AdminAPIHandlers) PutTrustDomainByName(echoCtx echo.Context, trustDomai
...
+		return chttp.LogAndRespondWithError(h.Logger, err, err.Error(), http.StatusNotFound)
+	}
+
+	err = h.Datastore.DeleteRelationship(ctx, relationship.ID.UUID)
+	if err != nil {
+		err = fmt.Errorf("failed getting relationships: %v", err)
+		return chttp.LogAndRespondWithError(h.Logger, err, err.Error(), http.StatusInternalServerError)
+	}
+
+	err = chttp.WriteResponse(echoCtx, http.StatusOK, "relationship deleted")
```
Victorblsilveira, line 405: Our responses should be in JSON rather than just strings. I am not sure how this message will be displayed to the end user, but to be consistent, should be something like `{ message: "relationship deleted" }`. Will also be good to use the response defined in the API specs.

pkg/server/endpoints/admin.go
mgbcaio: The `findRelationshipByID` already return the error wrapped up and log it. Doing this would result in 2 error messages.

pkg/server/endpoints/admin.go
mgbcaio: same as above

maxlambrecht (approved):

PR #231, opened 2023-06-19
Allow the customer to optionally set the `created_at` field . Improve datastore tests.

maxlambrecht:
**Pull request check list**

- [ ] Documentation updated?

**Affected functionality**

Allowing the customer to optionally set the `created_at` field while using `now()` as the default value when not provided, providing flexibility for specific use cases where the customer wants to control the creation timestamp. This can be also useful for testing. 

On the other hand, the Datastore layer continues not allowing the customer to set the `updated_at` field which aligns with the typical practice of using the current timestamp as the indicator of when a record was last modified. This ensures the integrity of the auditing process, as it prevents tampering with the timestamp for auditing purposes.

**Description of change**

Enabled setting the `created_at` field when inserting a new entity (`TrustDomain`, `Relationship`, `JoinToken`, `Bundle`).

Improved the Datastore tests to enhance consistency, streamline Datastore instance cleanup, and make various minor improvements.

### DIFF
--- a/pkg/server/db/postgres/bundles.sql.go
+++ b/pkg/server/db/postgres/bundles.sql.go
@@ -7,13 +7,14 @@ package postgres
 
 import (
 	"context"
+	"time"
 
 	"github.com/jackc/pgtype"
 )
 
 const createBundle = `-- name: CreateBundle :one
-INSERT INTO bundles(data, digest, signature, signing_certificate, trust_domain_id)
-VALUES ($1, $2, $3, $4, $5)
+INSERT INTO bundles(data, digest, signature, signing_certificate, trust_domain_id, created_at)
+VALUES ($1, $2, $3, $4, $5, $6)
 RETURNING id, trust_domain_id, data, digest, signature, signing_certificate, created_at, updated_at
 `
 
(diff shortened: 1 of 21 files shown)


Victorblsilveira (commented):
The changes are OK, but I don't understand the real necessity of allowing the user to control the `created_at`. Can you explore more this use case? Also, do you think we need to create some validations regarding the timestamp? Such as the time cannot be in the future?

maxlambrecht:
> The changes are OK, but I don't understand the real necessity of allowing the user to control the `created_at`. Can you explore more this use case? Also, do you think we need to create some validations regarding the timestamp? Such as the time cannot be in the future?

The most obvious use-case for me and the reason I'm adding this flexibility is for testing purposes. For example, you added a test that creates a `TrustDomain` setting the `created_at` expecting that it would be stored in the DB with the time you set:

```
td := &entity.TrustDomain{
              Name:      spiffeid.RequireTrustDomainFromString(tdName),
              CreatedAt: time.Now().Add(time.Duration(i+1) * time.Minute),
	}
td = createTrustDomain(ctx, t, ds, td)
```
Well, currently this is not happening, it is always set as the default `now()`, which _feels_ kind of surprising. 

In general the business layer won't deal with the `created_at` and just pass empty, and the Datastore layer will set it as the default which is `now()`. On the other hand, for testing purposes, I want to set it as anything I want, even in the future, like you were trying in your code.

Victorblsilveira (approved):

